PR #71, opened 2021-06-22
Optimize claimXpRewards

brian-man:
Using `solc.exe --gas --optimize cryptoblades.sol`, 24234 estimated gas savings.

### DIFF
--- a/contracts/cryptoblades.sol
+++ b/contracts/cryptoblades.sol
@@ -519,10 +519,11 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         // they meditate on what they've learned
         for(uint256 i = 0; i < characters.balanceOf(msg.sender); i++) {
             uint256 char = characters.tokenOfOwnerByIndex(msg.sender, i);
-            if(xpRewards[char] < 65536)
-                characters.gainXp(char, uint16(xpRewards[char]));
-            else
-                characters.gainXp(char, 65535);
+            uint256 xpRewardsToClaim = xpRewards[char];
+            if (xpRewardsToClaim > 65535) {
+                xpRewardsToClaim = 65535;
+            }
+            characters.gainXp(char, uint16(xpRewardsToClaim));
             xpRewards[char] = 0;
         }
     }


contracts/cryptoblades.sol
```diff
@@ -519,10 +519,11 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
             uint256 char = characters.tokenOfOwnerByIndex(msg.sender, i);
-            if(xpRewards[char] < 65536)
-                characters.gainXp(char, uint16(xpRewards[char]));
-            else
-                characters.gainXp(char, 65535);
+            uint256 xpRewardsToClaim = xpRewards[char];
+            if (xpRewardsToClaim > 65535) {
+                xpRewardsToClaim = 65535;
+            }
+            characters.gainXp(char, uint16(xpRewardsToClaim));
```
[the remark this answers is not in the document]
raymond-h: As I detailed in #74, it is unlikely to be an issue, but it is also good practice in general just in case when it might be an issue.

By nature of smart contract code and the blockchain, the concept of an atomic swap does not exist and is not needed anyway. It is simply enough to move `xpRewards[char] = 0` to just after `uint256 xpRewardsToClaim = xpRewards[char]`. See my suggested changes.

contracts/cryptoblades.sol
```diff
@@ -519,10 +519,11 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
             xpRewards[char] = 0;
```
raymond-h, line 527:
```suggestion
```

contracts/cryptoblades.sol
```diff
@@ -519,10 +519,11 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         // they meditate on what they've learned
         for(uint256 i = 0; i < characters.balanceOf(msg.sender); i++) {
             uint256 char = characters.tokenOfOwnerByIndex(msg.sender, i);
-            if(xpRewards[char] < 65536)
-                characters.gainXp(char, uint16(xpRewards[char]));
-            else
-                characters.gainXp(char, 65535);
+            uint256 xpRewardsToClaim = xpRewards[char];
```
raymond-h, line 522:
```suggestion
            uint256 xpRewardsToClaim = xpRewards[char];
            xpRewards[char] = 0;
```

raymond-h (changes requested):

brian-man:
I tried a dozen ways to get github to properly update this PR but something about the private-to-public conversion of the base repo seems to have broken github.

Updated changes now created here:
https://github.com/CryptoBlades/cryptoblades/pull/153

Issue #110, opened 2021-06-25
Claim flow


devinep52:
Claim should be moved to settings drop down

First promopt "Need Gas? Try our WAX Bridge, which will pay you .5% under market rate to sell your WAX for BNB!" Button to the wax bridge, button for Next

Next prompt, 
"If you stake your SKILL now, we will give you a 10% bonus in SKILL that you can use in-game right away!" Button to stake page
Under that, just small text that is actually a link, "No thanks, I'd rather [pay x% in taxes and withdraw and] forfeit my bonus]" depending on if any of their funds are taxable (they may have waited 15 days thus no tax). 
Next prompt
 You are about to [ pay x% in Tax for early withdrawal costing you Y SKILL] forfeit all bonus SKILL earnings for 3 days costing you Z skill], are you sure you wish to continue? This action cannot be undone"

sarimash:
Just want to update on where we are with this: the work in this ticket is done, pending implementation of the wax bridge/URL and the staking bonus. (Will need to update the routing with the wax bridge information)

I left out the "costing you Z skill" at the end because I thought that was referring to the skill gains over the next three days that wouldn't be predictable, but looking back at this is that just supposed to be the bonus 10% of the current claimable amount? If so I can add that blurb in as well.

seiyria:
@sarimash I believe the bonus is supposed to be 10% of the current skill. May as well slide that back in.

PR #125, opened 2021-06-27
Convert Options.vue to typescript, add withdrawal workflow

sarimash:
Options menu:
image
Prompt 1:
image
Prompt 2: (Note that this is with a hardcoded 14% tax and 0.1623 unclaimed SKILL)
image
Prompt 3:
image
On clicking "I am sure" Metamask prompt will open:
image
With the tax set to zero, the second prompt becomes:
image
and the third becomes:
image

### New Feature Submissions

Partially addresses #110
### PR Description

Pushing this for review, since it will surely need some style/best practices input :) 

I have the currentWithdrawalTax hardcoded in the vue for display purposes, so will need to retrieve that from Store when it is available, as well as set up the Wax Bridge router link. Also pending is removing the claim functionality from the ClaimRewardsBar, but I didn't want to break that until this was finished.


seiyria (approved):

seiyria:
waiting on the flow to be done but this looks good

Issue #134, opened 2021-06-28
reforging reforged items should give a reduced bonus to the new item


seiyria:
if you have a 4* with LB/4b/5b it should transfer to a new item at a rate of 50% rounded up (so if you have 5LB it would give 3LB, 1 -> 1 etc).

PR #145, opened 2021-06-29
reforging reforged items should give a reduced bonus to the new item

kcper:
### All Submissions

https://user-images.githubusercontent.com/9114200/123780445-7d7d6100-d8d3-11eb-935d-c4f7a58dd70a.mp4

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/134
### Notes

### PR Description

Reforging now upgrades a weapon how it did originally and then carries over 50% LBs of burned weapon.
Hopefully this way of doing calculations in solidity is correct. 
Second reforge shows that 1/x LBs ara carried over as 1.


seiyria (approved):

contracts/weapons.sol
```diff
@@ -416,6 +421,10 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable,
             if(wbp.fiveStarBurnPoints > 10)
                 wbp.fiveStarBurnPoints = 10;
         }
+        wbp.lowStarBurnPoints = uint8(carriedLowStarBurnPoints.add(wbp.lowStarBurnPoints));
+        wbp.fourStarBurnPoints = uint8(carriedFourStarBurnPoints.add(wbp.fourStarBurnPoints));
+        wbp.fiveStarBurnPoints = uint8(carriedFiveStarBurnPoints.add(wbp.fiveStarBurnPoints));
+
```
raymond-h, line 427: Since the code that clamps each value to the cap is all done before this, it is entirely possible for each of these bonuses to exceed the cap. This is probably not what we want.

raymond-h (changes requested):

contracts/weapons.sol
kcper: Fixed. I was thinking for a while how to do that the best, either add another ton of ifs for checking if each carried bonus is > 0 and then if new bonus does not exceed the limit or just check for everything by default every time we do a reforge. The later one seems like a better option.

raymond-h (approved):
Approved, will be merged later during a more opportune moment so that it does not block frontend-only changes from being deployed.

PR #124, opened 2021-06-26
Add contract support for burn dust

brian-man:
This change adds contract support for the introduction of "burn dust".  Existing functionality related to reforging weapons is unchanged.

To weapons.sol:
- add `Burned` event
- add `ReforgedWithDust` event
- add `getDustSupply()`
- add `burn()`
- add `reforgeWithDust()`

To cryptoblades.sol:
- add `burnWeapon()`
- add `reforgeWeaponWithDust()`
- add `setBurnWeaponValue()`
- add `setReforgeWeaponDustValue()`

Related to suggestion:
https://cryptoblades.hellonext.co/b/Suggestions/p/change-burn-of-reforge-into-a-claimable-item

PR #154, opened 2021-06-30
Add contract support for burn dust

brian-man:
This change adds contract support for the introduction of "burn dust".  Existing functionality related to reforging weapons is unchanged.

To weapons.sol:
- add `Burned` event
- add `ReforgedWithDust` event
- add `getDustSupply()`
- add `burn()`
- add `reforgeWithDust()`

To cryptoblades.sol:
- add `burnWeapon()`
- add `reforgeWeaponWithDust()`
- add `setBurnWeaponValue()`
- add `setReforgeWeaponDustValue()`

Related to suggestion:
https://cryptoblades.hellonext.co/b/Suggestions/p/change-burn-of-reforge-into-a-claimable-item


brian-man:
Github is broken. This PR replaces https://github.com/CryptoBlades/cryptoblades/pull/124.

poshdan:
I'd like to have a discussion on the way we store these game currencies (on the contract level) because it affects our options for the future (and the size of each contract, if we're adding state variables).
I think dust makes a good candidate for trading, as it saves on the headache of listing dozens of weapons and buying them one by one. We need to consider this new currency and any future currencies for our market.
ERC20 comes to mind, though it may be a little too much (requiring approval, for example) for currencies that are only intended for use in game. We could design our own lite interface for it too. I'd like to hear some opinions on which direction to go forward.


Also a minor note about the burn function being added to the weapons contract, it's more than just a public wrap of the private _burn function (and it is named after a standard function of the NFT interface), so I thought it may be best to use something more descriptive like "burnToDust" to avoid coding accidents in the future (example would be a mechanic that's meant to destroy swords for another purpose, without giving dust)

brian-man:
> > I think dust makes a good candidate for trading, as it saves on the headache of listing dozens of weapons and buying them one by one. We need to consider this new currency and any future currencies for our market.
I would like to see this happen as well.  Was going to raise the suggestion as we get further along.

My thinking was that the unclaimed skill model is a good one w/r/t allowing users to defer the claiming costs to whenever they want.  Similar could be applied for dust.  That said, if it's cheaper (gas) to have the burn function consume CBW and produce CBD (CryptoBlades Dust) in a single transaction then that's simpler.

brian-man:
> Also a minor note about the burn function being added to the weapons contract, it's more than just a public wrap of the private _burn function (and it is named after a standard function of the NFT interface),

Yah, I noticed _mint and _burn are standardized rather than something CB-specific.  As such, changing the name for the functions I'm adding probably makes sense.

crypto-smoke:
> I think dust makes a good candidate for trading, as it saves on the headache of listing dozens of weapons and buying them one by one. We need to consider this new currency and any future currencies for our market.
> ERC20 comes to mind, though it may be a little too much (requiring approval, for example) for currencies that are only intended for use in game. We could design our own lite interface for it too. I'd like to hear some opinions on which direction to go forward.

I think staying with ERC20 is a good way to go for dust. It allows you to trade it outside the game if desired, and enables trading without relying on the CB interface. As for approvals, just set the default approval amount on the first transaction with dust to 0xFF.. like exchanges do and they'll never have to approve that contract again. 

Also suggest "iron dust" or something for a name.

raymond-h:
I do personally like the idea of using ERC20 for dust. The main contract is getting very bloated, so I think involving more contracts is something we do want to do - and at that point, there is not much reason not to just go the whole way and use ERC20. In line with what @crypto-smoke mentioned, OpenZeppelin's implementation of the ERC20 standard would even allow us to have users automatically approve a given contract with a stupidly high number and thus avoid the issue of approval.

Another option I recently researched and grew a liking to is using the ERC777 standard, which is backwards-compatible with ERC20 but adds several goodies:
* *Operators* are special addresses that always have full control over some specific token on behalf of a user - basically as if they had unlimited allowance for that user's tokens all the time. This would thus be another alternative for users to not need to constantly approve transactions.
* *Receive hooks* plus support for sending tokens together with *arbitrary data*. This is yet another alternative for the issue of approval - the user could instead, in a single transaction, just send the dust token to the game contract together with some extra data specifying the operation they wanna do, like mint a new weapon or whatever. If multiple ERC777-based resources are needed this wouldn't work though.

If we choose to have a concept of "unclaimed" dust, ERC777 could also let us nicely support having users actually transfer tokens *from* their wallet *into* the contract too, thanks again to the receive hooks that lets the receiving end know who is sending the tokens.

I think an ERC777 token for dust, with the game contract set as a default operator, could be really neat. Any objections?

raymond-h:
Other than the question of whether to keep dust as part of the game contract or as a separate contract, this PR looks good to me.

poshdan:
My main concern with a token interface was the additional gas cost. Granted, if we have the claim system for these types of resources, most of that is negated.

I'm leaning towards having future claimable resources in a separate contract. Who knows what else we'll add eventually (even if layer 2 stuff doesnt backwash to our contracts). I'd rather it didn't pile up to a point where we hit some contract size limit.

crypto-smoke:
My vote goes to a seperate token. I'm starting to get a bit overwhelmed with the 47 different versions of SKILL that exist in the contract, of which only one is actually transferable... 

ERC20 or 777 is fine as long as you're comfortable with it. It might be worth exploring 777 with this for now as it sounds like an excellent base for future currencies as well.

brian-man:
> My main concern with a token interface was the additional gas cost. Granted, if we have the claim system for these types of resources, most of that is negated.
> 
> I'm leaning towards having future claimable resources in a separate contract. Who knows what else we'll add eventually (even if layer 2 stuff doesnt backwash to our contracts). I'd rather it didn't pile up to a point where we hit some contract size limit.
How do you feel about putting claimables on the NFT itself?  This means that the cost of each claimable is carried by the token rather than piling them all in one contract.

poshdan:
> How do you feel about putting claimables on the NFT itself? This means that the cost of each claimable is carried by the token rather than piling them all in one contract.

I think that's pretty good actually. Then we're never limited at all.

brian-man:
@seiyria had some concerns about keeping 5star dust special.

Option A: 1<<star_count
1star = 1<<1 = 2
2star = 1<<2 == 4
3star = 1<<3 == 8
4star = 1<<4 == 16
5star = 1<<5 == 32

Option B: 1<<(2*star_count)
1star = 1<<2 = 4
2star = 1<<4 == 16
3star = 1<<6 == 64
4star = 1<<8 == 256
5star = 1<<10 == 1024

Option C: 1<<(1<<star_count))
1star = 1<<2 = 2
2star = 1<<4== 16
3star = 1<<8 == 256
4star = 1<<16 == 65536
5star = 1<<32 == 4billion

I think at least option C is good enough to make 5star special but allow a single Dust currency.

If we went with a mathematical approach like this to separating star value for dust, then we don't have to change anything scary when introducing dust as a token.  Nor do we have to introduce 5 tokens.

---

Some additional color:
TLDR: look at C 4star vs. 5star.

The numbers there are clear: 1 and 2 are in the same game.  Upgrading 1-2 to 3 is spendy.  Upgrading 4star is kinda ludicrous.  Upgrading to 5star is basically "no".

So you effectively use those shift amounts to pack the dust balance into a uint256 for the unclaim (like I have in the PR but with different shift values).  When you go token, you just use the number directly.

A 5star is basically 1<<14 dust tokens.  (=1<<32 / 1<<18)
A 4star is 0.25 dust tokens.
A 3star is 0.0009765625 dust tokens.

---

One range of numbers (C-1):
A 5star is basically 1<<14 dust tokens.  (=1<<32 / 1<<18)
A 4star is 0.25 dust tokens.
A 3star is 0.0009765625 dust tokens.

---

Another range of numbers (C-2):
A 5star is basically 1<<16 dust tokens.  (=1<<32 / 1<<18)
A 4star is 1 dust token.
A 3star is 0.00390625 dust tokens.

brian-man:
From discussion with Seiyria:

So you're thinking something like this....
0x100000000000010001

If I try to spend
0x000001000000000000
I can't.

But I can spend any of
0x100000000000000000
0x000000000000010000
0x000000000000000001

You can still fit that in a single BEP-20.  It's just a little weird to use it that way.

brian-man:
Latest from dev channel with @seiyria :
PR 1 = [a] dust from stars adjusted [b] double til 50 for 1/2/3 star during normal reforge (no dust involved)
PR 2 (this PR, with adjustments) = unclaimed dust introduction
PR 3 = reforge allows specifying quantity of unclaimed dust to apply
PR 4 = mass burn to unclaimed in contract
PR 5 = unclaimed dust -> reforge UI
PR 6 = single burn UI
PR 7 = multi-select burn UI

raymond-h (changes requested):
Based on the discussion on Discord, it was established that there were several changes that should be done, right? Hence I'll mark this PR as "changes requested" for now then. Please get back to me if that's incorrect.

raymond-h (approved):

PR #163, opened 2021-06-30
Issue 135 - display xp payout on fights

Nizzypal:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

(insert description here)

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -84,7 +84,7 @@
           </div>
 
           <div class="row mb-3" v-if="targets.length > 0">
-            <div class="col-md-3 col-sm-12 encounter text-center d-flex flex-column justify-content-center" v-for="(e, i) in targets" :key="i">
+            <div class="col-md-3 col-sm-12 col-xs-12 encounter text-center d-flex flex-column justify-content-center" v-for="(e, i) in targets" :key="i">
               <img class="mr-auto ml-auto" :src="getEnemyArt(e.power)" alt="Enemy">
 
               <div class="encounter-element">
@@ -386,13 +386,7 @@ export default {
   font-size: 2em;
 }
 
-@media only screen and (max-width: 900px) {
-
-  div.encounter.text-center{
-    flex-basis: 0 !important;
-    width: 100%;
-  }
-
+div.encounter.text-center{
+  flex-basis: auto !important;
 }
-
 </style>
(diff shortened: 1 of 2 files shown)


frontend/src/views/Combat.vue
```diff
@@ -137,6 +140,7 @@ export default {
       intervalMinutes: null,
       timeSeconds: null,
       timeMinutes: null,
+      fightXpGain: 32,
```
seiyria, line 143: XP is not guaranteed to be 32; this needs to be correctly calculated - if I remember right it's from 14-17, but that could be clarified.

frontend/src/views/Combat.vue
```diff
@@ -286,10 +290,24 @@ export default {
         this.error = e.message;
       }
     },
+
     formattedSkill(skill) {
       const skillBalance = Web3.utils.fromWei(skill, 'ether');
       return `${new BN(skillBalance).toFixed(6)} SKILL`;
     },
+
+    onChooseWeapon(targetToFight) {
```
seiyria, line 299: this function is confusingly named

frontend/src/views/Combat.vue
Nizzypal: who should we ask... i saw that number in cryptoblades.sol tho it looks like its some sort  of  special case
```
    function migrateTo_1ee400a() public {
        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Not admin");

        fightXpGain = 32;
    }
```

frontend/src/views/Combat.vue
Nizzypal: right, yeah I did thename when the function was still an  event handler for an emit... it should be  changed.

seiyria:
So if I were doing this, I would create a bold yellow text, position it absolutely over the art but near the bottom (above the button), instead of adding another prop to buttons.

We'll also need to get the exact amount of XP needed for this, but that you'll probably need some help getting as I'm not entirely sure how it's done yet.

Nizzypal:
> So if I were doing this, I would create a bold yellow text, position it absolutely over the art but near the bottom (above the button), instead of adding another prop to buttons.
> 
> We'll also need to get the exact amount of XP needed for this, but that you'll probably need some help getting as I'm not entirely sure how it's done yet.

who do you think can we ask reg the exp?

seiyria (approved):

PR #184, opened 2021-07-06
Issue 107 - market improvement: sort by price high->low and low->high

Nizzypal:
#184 

### New Feature Submissions

### Notes

### DIFF
--- a/frontend/src/views/Market.vue
+++ b/frontend/src/views/Market.vue
@@ -39,7 +39,6 @@
                 v-if="activeType === 'weapon'"
                 :showGivenWeaponIds="true"
                 :weaponIds="allSearchResults"
-                :sortAscending="false"
                 :showLimit="60"
                 :showReforgedToggle="false"
                 :canFavorite="false"
@@ -175,7 +174,6 @@
                 :showReforgedToggle="false"
                 :canFavorite="false"
                 :weaponIds="searchResults"
-                :sortAscending="false"
                 @change="onSortChange"
                 v-model="selectedNftId">
 
@@ -446,13 +444,15 @@ export default Vue.extend({
     },
 
     onSortChange( e: any) {
+      const sort = e === 'Ascending Price' ? 'Ascending' : 'Descending';
+
       if (e === 'Ascending Price') {
-        this.allSearchResults = this.sortList(this.nftPricesById,'Ascending');
-        this.searchResults = this.sortList(this.nftPricesById,'Ascending');
+        this.allSearchResults = this.sortList(this.nftPricesById,sort);
+        this.searchResults = this.sortList(this.nftPricesById,sort);
       }
       else {
-        this.allSearchResults = this.sortList(this.nftPricesById,'Descending');
-        this.searchResults = this.sortList(this.nftPricesById,'Descending');
+        this.allSearchResults = this.sortList(this.nftPricesById,sort);
+        this.searchResults = this.sortList(this.nftPricesById,sort);
       }
     },
 
@@ -499,11 +499,11 @@ export default Vue.extend({
 
     },
 
-    sortList( List:  Record<any, any>, direction: string): any {
+    sortList( list:  Record<any, any>, direction: string): any {
       let temp: string[];
       if (direction.includes('Ascending'))
-        temp = Object.keys(List).sort(function(a,b){return List[parseInt(a,10)] - List[parseInt(b,10)];});
-      else temp = Object.keys(List).sort(function(a,b){return List[parseInt(b,10)] - List[parseInt(a,10)];});
+        temp = Object.keys(list).sort((a, b) => list[parseInt(a,10)] - list[parseInt(b,10)]);
+      else temp = Object.keys(list).sort((a, b) => list[parseInt(b,10)] - list[parseInt(a,10)]);
       return temp;
     },
 
--- a/frontend/src/views/Market.vue
+++ b/frontend/src/views/Market.vue
@@ -39,9 +39,11 @@
                 v-if="activeType === 'weapon'"
                 :showGivenWeaponIds="true"
                 :weaponIds="allSearchResults"
+                :sortAscending="false"
                 :showLimit="60"
                 :showReforgedToggle="false"
                 :canFavorite="false"
+                @change="onSortChange"
                 v-model="selectedNftId">
 
                 <template #above="{ weapon: { id } }">
@@ -173,6 +175,8 @@
                 :showReforgedToggle="false"
                 :canFavorite="false"
                 :weaponIds="searchResults"
+                :sortAscending="false"
+                @change="onSortChange"
                 v-model="selectedNftId">
 
                 <template #above="{ weapon: { id } }">
@@ -441,6 +445,17 @@ export default Vue.extend({
       this.nftPricesById = {};
     },
 
+    onSortChange( e: any) {
+      if (e === 'Ascending Price') {
+        this.allSearchResults = this.sortList(this.nftPricesById,'Ascending');
+        this.searchResults = this.sortList(this.nftPricesById,'Ascending');
+      }
+      else {
+        this.allSearchResults = this.sortList(this.nftPricesById,'Descending');
+        this.searchResults = this.sortList(this.nftPricesById,'Descending');
+      }
+    },
+
     async loadMarketTaxes() {
       if(!this.characterMarketTax) {
         const tax = await this.getMarketTax(this.Characters.options.address) as string;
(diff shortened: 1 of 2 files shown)


kcper:
Keep in mind that:
- right now this will sort just the 40 characters / 60 weapons that are visible
- when pagination gets merged, this will sort only the things visible on the current page which is confusing, instead of sorting all the listings which is desired as a result of this issue

seiyria:
@kcper can you make this mesh with your PR? it'll be important to have this with pagniation

frontend/src/views/Market.vue
```diff
@@ -441,6 +445,17 @@ export default Vue.extend({
       this.nftPricesById = {};
     },
 
+    onSortChange( e: any) {
+      if (e === 'Ascending Price') {
```
seiyria, line 449: this can be cleaned up by doing `const sort = e === 'Ascending Price' ? 'Ascending' : 'Descending'` then reusing that variable.

frontend/src/views/Market.vue
```diff
@@ -478,10 +493,18 @@ export default Vue.extend({
...
       await Promise.all(nftIds.map(async nftId => {
         const price = (await this.lookupNftPrice(nftId))!;
-
         void price;
         this.nftPricesById[nftId] = price;
       }));
+
+    },
+
+    sortList( List:  Record<any, any>, direction: string): any {
```
seiyria, line 502: please don't use an uppercase variable name; it can be confused with an enum or constant

frontend/src/views/Market.vue
```diff
@@ -478,10 +493,18 @@ export default Vue.extend({
...
+      let temp: string[];
+      if (direction.includes('Ascending'))
+        temp = Object.keys(List).sort(function(a,b){return List[parseInt(a,10)] - List[parseInt(b,10)];});
```
seiyria, line 505: this can be `(a, b) => list[+a] - list[+b]` - much more concise

seiyria (changes requested):

kcper:
> @kcper can you make this mesh with your PR? it'll be important to have this with pagniation

Sorry what do you mean? For market grid the filtering and sorting should be done in the backend, there's no use of having it done at the frontend with pagination. This kind of sorting can be implemented for non-market weapon/character grids.

Nizzypal:
Ok gotcha.

seiyria:
I wasn't sure if there was anything you could do with solidity to make it work. If not, that might be fine too.

Nizzypal:
Since I didnt make  the other contracts,I dont want to assume what impact on gas fees sorting in that layer would have.  There might be  too many interactions.  And  im still learning solidity.

kcper:
As far as I know reading from contract is free so I wouldn't worry about fees, because it's not a transaction.

We need to do the sorting in the backend (right now we only have the contract for it so no other choice if we want to have it now), because only this way we can make sure the pages are formed in the sorting order.

Just to clarify - when the pagination gets merged, this solution will only sort the things on the current page so e.g.:
- Page 1 prices: 10, 20, 30, 40
- Page 2 prices: 3, 5, 10, 30

While it should be:
- Page 1 prices: 3, 5, 10, 10
- Page 2 prices: 20, 30, 30, 40

Nizzypal:
@kcper I see.  Where in the  front end do you dump your paginated data?  I  can collect all of it and then sort.

kcper:
Pagination doesn't dump all the data, just the data on current page, so you won't be able to do that, because you will never have all data in the frontend. Pagination is not only there to let one browse all listings, but also to reduce the amount of listings you load at a time to the amount on one page.

The thing you mentioned about tables with sorted data is actually one of the implementations I proposed. Not sure if it's supposed to be in another contract or just as additional sorted arrays in the current one, but that's the best idea I had. Probably KR will make a final decision on that.

kcper:
Btw if @seiyria thinks the current state of your solution is useful then don't mind me, I'm just sharing my point of view. But this will most likely need to be reworked later.

Nizzypal:
@kcper i understand,  Maybe I can do a workaround since the user won't know the items not shown until he/she asks for it, I can maybe  do some sort of "lazy sorting" (sorting as you give the data)... so for example, he  just views page 1,  data for page 1  would be sorted.  If he  views page 2, then  data for page 1 and 2 would be sorted and so  forth and so on. 
In that light, could you please tell me  where you  give your paginated data in the front end?

seiyria:
It's okay, but I'd like as little rework as possible ultimately. Rework can fall through the cracks if we're not careful.

Nizzypal:
> Btw if @seiyria thinks the current state of your solution is useful then don't mind me, I'm just sharing my point of view. But this will most likely need to be reworked later.

Actually there are merits for both approaches, but  from my end, what I can do is try to merge what you did with what I did.   Hence  it would  help alot if you could tell me the entry point for your data in the frontend :)

kcper:
Still not possible, how do you imagine it? User sees some weapons at page 1, then they go to page 2 and then they see same weapons again, because that's how they are sorted now?

Regarding the results - they are stored at the same variable as right now, check out the changes to Market.vue here https://github.com/CryptoBlades/cryptoblades/pull/155/files

Nizzypal:
> Still not possible, how do you imagine it? User sees some weapons at page 1, then they go to page 2 and then they see same weapons again, because that's how they are sorted now?
> 
> Regarding the results - they are stored at the same variable as right now, check out the changes to Market.vue here https://github.com/CryptoBlades/cryptoblades/pull/155/files

It will still be sorted in the context of  all the data you have passed in the front  end, at least that's the idea.  I'll see when I try it lolz.   Thanks for the info.

kcper:
> > Still not possible, how do you imagine it? User sees some weapons at page 1, then they go to page 2 and then they see same weapons again, because that's how they are sorted now?
> > Regarding the results - they are stored at the same variable as right now, check out the changes to Market.vue here https://github.com/CryptoBlades/cryptoblades/pull/155/files
> 
> It will still be sorted in the context of all the data you have passed in the front end, at least that's the idea. I'll see when I try it lolz. Thanks for the info.

Yeah try what you want, but it can not work with how pagination is currently done, at least doesn't make sense to me.





> > Btw if @seiyria thinks the current state of your solution is useful then don't mind me, I'm just sharing my point of view. But this will most likely need to be reworked later.
> 
> Actually there are merits for both approaches, but from my end (since I can't make design decisions), what I can do is try to merge what you did with what I did. Hence it would help alot if you could tell me the entry point for your data in the frontend :)

Also that's why I mentioned under this issue that it's better to wait for pagination merge to avoid conflicts and started a conversation about ideas for the implementation.

Nizzypal:
> > > Still not possible, how do you imagine it? User sees some weapons at page 1, then they go to page 2 and then they see same weapons again, because that's how they are sorted now?
> > > Regarding the results - they are stored at the same variable as right now, check out the changes to Market.vue here https://github.com/CryptoBlades/cryptoblades/pull/155/files
> > 
> > 
> > It will still be sorted in the context of all the data you have passed in the front end, at least that's the idea. I'll see when I try it lolz. Thanks for the info.
> 
> Yeah try what you want, but it can not work with how pagination is currently done, at least doesn't make sense to me.
> 
> > > Btw if @seiyria thinks the current state of your solution is useful then don't mind me, I'm just sharing my point of view. But this will most likely need to be reworked later.
> > 
> > 
> > Actually there are merits for both approaches, but from my end (since I can't make design decisions), what I can do is try to merge what you did with what I did. Hence it would help alot if you could tell me the entry point for your data in the frontend :)
> 
> Also that's why I mentioned under this issue that it's better to wait for pagination merge to avoid conflicts and started a conversation about ideas for the implementation.

Cool, hey I have some questions about the code you're about to merge.  Can I get your handle at discord so I'll know who to ask :)

We can just try to merge what you did and what I did and just adjust the sorting to take what you did in paginnatioin.   i think it can be done.

kcper:
My discord is Taloner.

Nizzypal:
Kcper and I talked about how his paging is implemented.  I believe I can make this work as soon as his code  is  visible  & accessible to me.  @kcper @seiyria

kcper:
> Kcper and I talked about how his paging is implemented. I believe I can make this work as soon as his code is visible & accessible to me. @kcper @seiyria

I'd suggest first explaining in detail how you see it working, get @seiyria approval and follow up with the implementation. I don't see it working like you explained so far to be honest.

Nizzypal:
Sure @kcper.   So for each page of data you send, ill get some prop or  variable (probably an array of strings) to copy the data, so taht when  the user changes pages, the past pages would  still be  in that variable the sorting  would  always include all data  in that variable thus far.
So using your example, if the user looksat the 1st page, it will sort as you said it would, but then when  he looks  at the second page, he will see the variables in the first page as he should since the sorting would include both the second page and first page of data you sent because they will all be in that variable

Nizzypal:
If you can send me your  code  that would get the data, i can check if it works  (which incidentally would  take care of some  of  ht  merging that will be done)

kcper:
It won't work like that, it can't work like that. If I didn't manage to explain it to you until now then you do you and see what happens when you implement it. My code is in my pull request I linked.

Nizzypal:
> It won't work like that, it can't work like that. If I didn't manage to explain it to you until now then you do you and see what happens when you implement it. My code is in my pull request I linked.

As long  as your data  flows thru the front end, in theory it can be  capured and  used  for sorting.  Its a basic premise.  And why do you say it won't  work?   Have  you tried it?

seiyria:
I'm going to close this for now, since the approach is entirely different.

PR #185, opened 2021-07-06
Remove 'I understand/continue' step from onboarding

crypto-smoke:
Apeswap has officially listed SKILL so there's no more spooky warning.


seiyria (approved):

seiyria:
we're listed now, so we don't need this anymore.

Issue #190, opened 2021-07-08
Allow private trades on market


crypto-smoke:
Some users want to sell only to a specific other user. Allow listing of items to only be purchased by a specific address. This will keep users safe by still going through the marketplace but they wont have to risk listing an item and having a third party grab it before the desired recipient can.

seiyria:
if we do this, it does have to do tax like the market does - probably the same tax value. I'm in favor of this.

kcper:
Potential implementation:
- listing an item for some specific address is straight forward, just add another input below the price for the address (fill with "optional" and tooltip explaining what it does
- for buying we might let the buyer just search the ID which only they will be able to see or get another tab like "Offers" with listings just for the buyer (first one probably better for now right?)

crypto-smoke:
Anyone could search for ID's for things that are up for private sale, so that wouldn't work out well. If it's locked so only a specific address can buy then there's no way anyone other than the intended recipient could purchase it. 

And yeah I failed to mention in the initial comment but the tax should absolutely remain in place for these transactions 👍

kcper:
So basically add something like targetBuyer property to the listing and filter the listings with that property specified on the frontend (only issue with that is that someone could buy the listing by finding it on chain and writing to the contract from bscscan or whatever). Otherwise we'd need to add logic to the contract function which is getting quite complicated already after pagination gets merged. How does that sound @seiyria?

seiyria:
I think that's reasonable - specifying who its for and blocking purchase if that field exists makes sense. On top of that, it's minimal frontend work (imo) comparatively. You'll need a new field when listing (target buyer), and maybe a new filter ("for me"). I am in favor.

kcper:
Sounds good.

crypto-smoke:
> 
> 
> So basically add something like targetBuyer property to the listing and filter the listings with that property specified on the frontend (only issue with that is that someone could buy the listing by finding it on chain and writing to the contract from bscscan or whatever). 

Nope, if the contract finds that targetBuyer is not 0x00, then the transaction fails if the buyer address is not the one specified. Yes, you could find every private sale by interacting with the contract (or reading contract state direct from memory) but it wont matter because they cant fake being the correct buyer.

kcper:
Ok I'll implement the handling of that in the contract, too.

brian-man:
@kcper 
Recommendations (in addition to the new optional param on addListing):
[...]

PR #194, opened 2021-07-09
Allow private trades on market

kcper:
### All Submissions

Can't post vidoes with size > 10MB now...
https://discord.com/channels/833038505626173481/851805599628394507/862993561180307456
https://discord.com/channels/833038505626173481/851805599628394507/862993699557736448

Also added this in 3rd commit (not on videos):
image

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/190
### Notes

### PR Description

- added target buyer input in listing setup (marked as optional for non-private sale)
- added change target buyer button in "Search my weapons/characters" view
- if someone finds a listing on chain and tries to buy it, the transaction will not go through, because of this modifier:
```
modifier userAllowedToPurchase(IERC721 _tokenAddress, uint256 id) {
        require(listings[address(_tokenAddress)][id].targetBuyer == address(0) || listings[address(_tokenAddress)][id].targetBuyer == msg.sender, "Not target buyer");
        _;
    }
```


kcper:
Not sure why the build fails, any help appreciated.

seiyria:
UI-wise, I think this looks great.

frontend/src/views/Market.vue
```diff
@@ -531,6 +591,28 @@ export default Vue.extend({
...
+
+      const results = await this.changeMarketListingTargetBuyer({
+        nftContractAddr: this.contractAddress,
+        tokenId: this.selectedNftId,
+        newTargetBuyer: targetBuyer
+      });
+
+      this.selectedNftId = null;
+      this.waitingMarketOutcome = false;
+      this.marketOutcome = 'Successfully changed target buyer for '
```
seiyria, line 612: you could use template strings for this (`) to make it cleaner, like you did above

kcper: Sure

poshdan:
I noticed this changes the market listing structs in the market contract.
I'm fairly certain that kind of change is not upgrade-compatible, but @raymond-h might know better.

In hindsight, we could've had some kind of identifier in there to begin with, to make it more extensible. I'm not sure what to suggest as an alternative. One possible solution is to have extension mappings for listings but it may be tricky to pack an address/nftID pair into a key. Feel free to tag me with suggestions on how to get around it, and I'll take a look at it.

raymond-h:
@poshdan Solidity supports multi-level mappings, which is even used for tracking listings themselves in the first place. And even if it didn't, it is generally considered acceptable to merge arbitrary values by ABI-encoding and hashing the result.

Adding a field to the existing struct is indeed an issue though, which I will mention in my review.

contracts/NFTMarket.sol
```diff
@@ -51,6 +51,7 @@ contract NFTMarket is
         address seller;
         uint256 price;
         //int128 usdTether; // this would be to "tether" price dynamically to our oracle
+        address targetBuyer;
```
raymond-h, line 54: Adding this field will almost certainly cause problems wrt upgradeability of the contract. Instead, there should be a new multi-level mapping, similar to `listings` below, exclusively for the `targetBuyer` of a particular listing.

frontend/src/views/Market.vue
```diff
@@ -527,8 +586,28 @@ export default Vue.extend({
...
-        +this.activeType+' '+results.nftID+' to '+this.convertWeiToSkill(results.newPrice)+' SKILL';
+      this.marketOutcome = `Successfully changed price for ${this.activeType} ${results.nftID} to ${this.convertWeiToSkill(results.newPrice)} SKILL`;
+    },
+
+    async updateNftListingTargetBuyer() {
+
+      this.marketOutcome = null;
+      if(this.selectedNftId === null) return;
+
+      let targetBuyer = await (this as any).$dialog.prompt({ title: `Sell ${this.activeType}`, text: 'Target Buyer Address (optional)' });
```
raymond-h, line 597: Instead of doing `this as any`, the definition in `shims-vue.d.ts` should in theory make sure `$dialog` exists as a field already on `this` (the type of `this.$dialog` as per said file *is* just `any` anyway, but that's another issue). Would it be possible to remove the `as any` cast here?

frontend/src/views/Market.vue
```diff
@@ -633,7 +713,7 @@ export default Vue.extend({
 
         this.searchResultsOwned = this.search === this.defaultAccount;
         this.waitingMarketOutcome = false;
-        this.searchResults = result;
+        this.searchResults = await this.filterOutTargetBuyers(result) as string[];
```
raymond-h, line 716: I assume this cast was added because the type for `filterOutTargetBuyers` was somehow not picked up. Did you try explicitly specifying the return type on the `filterOutTargetBuyers` method itself?

raymond-h (changes requested):

frontend/src/views/Market.vue
kcper: Tried it at first, it didn't let me. I just checked again why and it seems like 
`export default Vue.extend({` instead of `export default {` causes that issue. After trying to remove `Vue.extend(` part, `this` is broken in other scenarios, e.g. `this.contracts` - property contract does not exist on `this` component. Let me know if I should investigate that more and find a fix or leave as is.

frontend/src/views/Market.vue
kcper: Seems to work when I do `: Promise<string>`. I will get rid of the casts.

contracts/NFTMarket.sol
kcper: That's a good idea, I did not think about it!

kcper:
Build errors, because this branch was created from failing main branch, will that be a problem during merge?

raymond-h:
Yes, please do rebase it, thank you.

frontend/src/views/Market.vue
raymond-h: Going to try and get this merged today, so I'll take a look at it myself instead, thanks anyway.

raymond-h (changes requested):

raymond-h:
Don't worry about actually doing the changes, I'm gonna do them myself right now anyway (plus merge in changes from `main` to resolve the conflict).

raymond-h:
Actually on second though, merging the changes from `main` is kinda complicated because of changes to market searching (which uses an API) and I'm not terribly familiar with either those changes or this PR's frontend changes. @kcper It might be easier for you to sort out since you wrote the code, so could you take a look at doing that after all?

kcper:
This is finished now.
@seiyria I looked at your api changes - not sure if we're on the same page.

buyerAddress is the address of a person browsing the market thus buyerAddress will be always passed to the api request (for market GET requests). I'm expecting the response to only have such nftIds that were listed without targetBuyer specified (these are being set by default to address(0) as target buyer - 0x0000000000000000000000000000000000000000) or with targetBuyer = buyerAddress.

I haven't seen any logic like this implemented so there may have been some confusion. Correct me if I'm wrong of course.

seiyria:
> buyerAddress is the address of a person browsing the market thus buyerAddress will be always passed to the api request (for market GET requests). I'm expecting the response to only have such nftIds that were listed without targetBuyer specified (these are being set by default to address(0) as target buyer - 0x0000000000000000000000000000000000000000) or with targetBuyer = buyerAddress.

This was not my take. I had figured that you would have a new "my trades" area that only shows up if you have trades (ie, query when the market loads, then show this).

> A case when you search for your own listings (so searching by sellerAddress = you) also needs special handling - when sellerAddress = buyerAddress (you browsing the market) then don't filter out listings with not matching targetBuyer, because they are your listings.

Based on my idea above, there would never be this situation, as you'd pass them circumstantially (sellerAddress when you want to see all buy a seller, and buyerAddress when you want to see your trades).

I can adjust it as needed if that doesn't make sense, though, so let me know.

kcper:
Search for private trades with these buttons
image

seiyria:
UI-wise, this looks fine to me. The conflicts need to be addressed, then we can see how it goes in test if @raymond-h agrees that it looks fine.

newb23:
Closes #190

raymond-h (approved):

kcper:
Did not test it after merge as this thing still prevents adding listings:
```
if(addFee > 0) {
    skillToken.safeTransferFrom(msg.sender, address(this), usdToSkill(addFee));
}
```
Approving any amount of SKILL spend (even >1M) still fails with "transfer amount exceeds allowance".
I'd hold off with merging until it's fixed.

PR #203, opened 2021-07-10
Claim flow

sarimash:
### All Submissions

image
Left routes to /portal, right goes to next dialog
image
Button routes to /stake, the white warning goes to last dialog
image
Left closes, right initiates the metamask confirmation to withdraw skill

### New Feature Submissions

Addresses https://github.com/CryptoBlades/cryptoblades/issues/110 .
The only current issues I know of is that the tax rate is hardcoded to 14% (for testing reasons) rather than retrieving from store, and the old method of claiming SKILL (clicking the unclaimed reward text) still works


frontend/src/components/Options.vue
```diff
@@ -1,38 +1,107 @@
...
+<script lang="ts">
 import Events from '../events';
+import { mapActions, mapState } from 'vuex';
+import BN from 'bignumber.js';
+import Web3 from 'web3';
+import { Accessors } from 'vue/types/options';
+import { BootstrapVueIcons } from 'bootstrap-vue';
+import Vue from 'vue';
+
+Vue.use(BootstrapVueIcons);
```
seiyria, line 48: this is included globally

frontend/src/components/Options.vue
```diff
@@ -1,38 +1,107 @@
...
+  currentWithdrawTax: number;
+  directStakeBonusPercent: number;
+}
+
+export default Vue.extend({
   created() {
     this.showGraphics = localStorage.getItem('useGraphics') === 'true';
     this.hideRewards = localStorage.getItem('hideRewards') === 'true';
     this.hideAdvanced = localStorage.getItem('hideAdvanced') === 'true';
+    this.currentWithdrawTax = 0.14;
```
seiyria, line 70: is this able to be gotten dynamically, or is it just for display purposes?

frontend/src/components/Options.vue
```diff
@@ -56,8 +125,25 @@ export default {
...
     },
-  },
-};
+    async onClaimTokens() {
+      if(this.canClaimTokens) {
+        await this.claimTokenRewards();
+      }
+    },
+    async claimSkill(stage: number) {
+      if(stage === 0) {
```
seiyria, line 134: can this be an enum instead?

frontend/src/components/Options.vue
sarimash: Pulling from store now, thanks

frontend/src/components/Options.vue
sarimash: Done! It's working as is, but should I have declared the enum somewhere else or anything?

seiyria: since the enum is just for this page, having it only on this page is fine

raymond-h (approved):

PR #283, opened 2021-07-14
New sqrt payout curve

poshdan:
Adds the new payout curve as discussed by the math team,
comes a migration file to avoid brief payout anomalies after deploy (optional)
(may need better accuracy on the params for the setter functions)

Expected gas cost increase of fights: ~786 (very negligible, may cost a few cents over the course of a day)

### DIFF
--- a/contracts/cryptoblades.sol
+++ b/contracts/cryptoblades.sol
@@ -76,6 +76,15 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         stakeFromGameImpl = _stakeFromGame;
     }
 
+    function migrateTo_7dd2a56() external {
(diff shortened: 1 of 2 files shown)


raymond-h:
One, would it be possible to use the existing functions for setting those fight rewards fields instead of using a full-blown `migrateTo_[X]` function?

Two, can we establish 100% whether we want this migration or not before I merge it (since you said it's optional)?

Code itself looks good otherwise.

raymond-h (changes requested):
Marking this as "changes requested" so I can filter it out in the PR UI. See my comment above for actual notes.

raymond-h (approved):

PR #285, opened 2021-07-15
False "very likely win" fixed

kamda-cyrial:
getElementAdvantage(playerElement, enemyElement) returns 1, -1 or 0, and the percentage change due to this advantage is rather 7.5%, so by multiplying the return value of the function by 0.075, we solve the problem

### All Submissions

### New Feature Submissions

### Notes

### PR Description

(insert description here)


sarimash:
The 0.075 is outside of the parentheses, the way it currently is the multiplier is (1 + 0.075*(1+1))

PR #291, opened 2021-07-15
Add durability to weapons

brian-man:
This changes adds durability to weapons, similar to stamina for characters.

Importantly, durability is intended to limit a weapon's fight use to all 4 characters.  While the same maximum (200) could be used, the minimum needed is 1 durability per fight which maps to 20 durability.  A higher number can be used but it really should be a multiple of 20 (4 characters with 5 fights each per 16 hours).

I opted to use a mapping rather than adding a field to Weapon to align with the implementation for the transfer cooldown and to avoid potential issues with expanding Weapon.

Lastly, the need to apply a drain to two NFTs during a single operation (stamina for character and durability for weapon) must be done carefully to avoid draining one of them and failing a require when draining the other.

### DIFF
--- a/contracts/cryptoblades.sol
+++ b/contracts/cryptoblades.sol
@@ -43,6 +43,7 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
 
         characterLimit = 4;
         staminaCostFight = 40;
+        durabilityCostFight = 1;
         mintCharacterFee = ABDKMath64x64.divu(10, 1);//10 usd;
         fightRewardBaseline = ABDKMath64x64.divu(1, 100);//0.01 usd;
         fightRewardGasOffset = ABDKMath64x64.divu(8, 10);//0.8 usd;
@@ -120,6 +121,8 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
 
     mapping(address => uint256) private _rewardsClaimTaxTimerStart;
 
+    uint8 durabilityCostFight;
+
     IStakeFromGame public stakeFromGameImpl;
 
     event FightOutcome(address indexed owner, uint256 indexed character, uint256 weapon, uint32 target, uint24 playerRoll, uint24 enemyRoll, uint16 xpGain, uint256 skillGain);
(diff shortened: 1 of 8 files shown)


newb23:
I've confirmed locally that this does indeed, work. 

`Error: Error: [ethjs-query] while formatting outputs from RPC '{"value":{"code":-32603,"data":{"message":"VM Exception while processing transaction: revert Not enough durability!","code":-32000,"data":{"0x17acf23d104dbd016fc9fbfa7e5df783a74618893490d79a8e76e763f371a577":{"error":"revert","program_counter":1123,"return":"0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000164e6f7420656e6f756768206475726162696c6974792100000000000000000000","reason":"Not enough durability!"},"stack":"c: VM Exception while processing transaction: revert Not enough durability!\n at Function.c.fromResults (/home/omni/.nvm/versions/node/v16.3.0/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:4:192416)\n at w.processBlock (/home/omni/.nvm/versions/node/v16.3.0/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:42:50915)\n at runMicrotasks (<anonymous>)\n at processTicksAndRejections (node:internal/process/task_queues:96:5)","name":"c"}}}}`

raymond-h (approved):

raymond-h (changes requested):
So to be honest I forgot about it at the time, but surely this also requires some frontend work before it can get merged.

kcper:
Starting the frontend work on it.

kcper:
All durabilities loaded when game loads.
Used weapon durability updates after combat.
All weapon durabilities updated every 10s.

https://user-images.githubusercontent.com/9114200/125966531-fcc7d449-bef0-4200-af9a-ed76dc1fd792.mp4

seiyria (approved):

raymond-h (approved):

raymond-h (changes requested):
According to @kcper the frontend isn't actually ready quite yet to merge, so I'll mark this PR as "changes requested" so no one accidently merges it thinking it's fine just yet.

raymond-h (approved):

PR #297, opened 2021-07-16
Reforging lb change

brian-man:
The current reforging model is problematic for future changes intended related to dust.

This change does the following:
- set the burn value for 1-3 star weapons to match the star count
- apply the doubling bonus for 1-2 star weapons to 50 LB (matching 3 star weapons)


brian-man:
The first change was proposed by @smoke and supported by @seiyria.
The second change was proposed by @seiyria and solves the problem of 1, 2, and 3 star dust not being interchangeable. 

Lastly if you look at the individual commits you'll notice that the logic is simplified for readability and to save gas.  I didn't measure the gas savings.

raymond-h (approved):

raymond-h (approved):

PR #298, opened 2021-07-16
Remove transfer cooldown from weapons

brian-man:


brian-man:
It wasn't clear to me if it's as simple as just removing the plumbing, or if some of that plumbing has to be retained but neutered.  Please advise.

contracts/weapons.sol
```diff
@@ -95,8 +89,6 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable,
     int128 public powerMultPerPointPWR; // 0.2575% (+3%)
     int128 public powerMultPerPointMatching; // 0.2675% (+7%)
 
-    mapping(uint256 => uint256) public override lastTransferTimestamp;
```
raymond-h, line 98: State variables cannot be removed, ever. The `override` keyword should be fine to remove however.

contracts/weapons.sol
```diff
@@ -6,17 +6,13 @@ import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol"
...
-import "./interfaces/ITransferCooldownable.sol";
 
-contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable, ITransferCooldownable {
+contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
 
     using ABDKMath64x64 for int128;
     using ABDKMath64x64 for uint16;
 
     bytes32 public constant GAME_ADMIN = keccak256("GAME_ADMIN");
-    bytes32 public constant RECEIVE_DOES_NOT_SET_TRANSFER_TIMESTAMP = keccak256("RECEIVE_DOES_NOT_SET_TRANSFER_TIMESTAMP");
```
raymond-h, line 17: I would like to keep this constant for now, as it is used in `migrations/29_transfer_cooldown.js`. While that is an old migration, and this constant won't actually be applicable in any way after this PR is merged, I would like to leave all applied migration files untouched for now.

Eventually I would like to merge all the old applied migrations we have right now into a single file - I think that would be a good time to finally remove this constant.

raymond-h (changes requested):
Largely good, but the frontend part of the weapon transfer cooldown code still has to be removed as well.

contracts/weapons.sol
brian-man: Okies.
Is there some language sugar or comment you recommend to make it clear it's no longer to be used?

raymond-h: I guess use `// UNUSED; KEPT FOR UPGRADEABILITY PROXY COMPATIBILITY`

raymond-h (changes requested):
Looks good to me now, just needs frontend work now.

raymond-h (approved):

PR #302, opened 2021-07-16
Remove doubling bonus for LB reforging

brian-man:


brian-man:
Had a long discussion @poshdan to talk through the ramifications of the preceding LB change.  There was/is concern that the number of minted weapons for a full reforge were lowered significantly with that PR.  This PR both addresses that issue and makes the model even simpler.

poshdan (approved):

Issue #236, opened 2021-07-12
navbar needs to be cleaned up significantly


seiyria:
ingame skill and normal skill take up too much space. in game skill should really get moved.

additionally, some routes can probably get moved out of the nav bar because they're a bit less important, such as leaderboards.

details TBD

bilbolPrime:
For in game skill, I felt it's best to just merge them together and show the details in a tool tip. Saves the player some mind math in case they are using both in game and real skill to buy an item. I don't think there will be much rounding headaches.
![image](https://user-images.githubusercontent.com/25177987/126031267-78949d82-edaf-49d6-bea3-7ef684250559.png)

![image](https://user-images.githubusercontent.com/25177987/126030609-64292a96-c5dc-4338-b853-07d9928f01fc.png)
On Hover:
![image](https://user-images.githubusercontent.com/25177987/126030619-e10ce1ee-819c-4f11-898b-b2bb75c86578.png)
![image](https://user-images.githubusercontent.com/25177987/126031291-ea231037-23e5-419c-b536-d378760fb40e.png)

For the leaderboard, I just moved it to the options menu
![image](https://user-images.githubusercontent.com/25177987/126030941-4da4bfe5-0718-40b9-9ce9-4fade7e0ab64.png)
![image](https://user-images.githubusercontent.com/25177987/126031306-4a8b87bc-6664-45e8-acd8-527618fd0b61.png)

Code on branch: 236-navbar-clean-up
Let me know if you have any comments or need anything more before opening a PR

PR #304, opened 2021-07-17
236-navbar-clean-up

bilbolPrime:
### All Submissions

image
image
image
image
image
image
image

### New Feature Submissions

* [x] #236 

### Notes

### PR Description

For in game skill, I felt it's best to just merge them together and show the details in a tool tip. Saves the player some mind math in case they are using both in game and real skill to buy an item. I don't think there will be much rounding headaches.

Navbar: Moved leadership and portal to menu + added grouped items for the menu. Did not add the tagname to the portal because it it didn't have it in navbar so I figured not being tracked.

### DIFF
--- a/frontend/src/components/Options.vue
+++ b/frontend/src/components/Options.vue
@@ -16,7 +16,7 @@
 
         <b-dropdown-item @click="claimSkill(ClaimStage.WaxBridge)">Claim Skill </b-dropdown-item>
 
-        <b-dropdown-item @click.native="$router.push('leaderboard')">Leaderboard </b-dropdown-item>
+        <b-dropdown-item @click.native="$router.push('leaderboard')" class="gtag-link-others" tagname="leaderboard_screen">Leaderboard </b-dropdown-item>
 
         <b-dropdown-item href="https://cryptoblades.gitbook.io/wiki/" target="_blank">Wiki <b-icon scale="0.8" icon="question-circle"/></b-dropdown-item>
 
--- a/frontend/src/components/Options.vue
+++ b/frontend/src/components/Options.vue
@@ -16,6 +16,8 @@
 
         <b-dropdown-item @click="claimSkill(ClaimStage.WaxBridge)">Claim Skill </b-dropdown-item>
 
+        <b-dropdown-item @click.native="$router.push('leaderboard')">Leaderboard </b-dropdown-item>
+
         <b-dropdown-item href="https://cryptoblades.gitbook.io/wiki/" target="_blank">Wiki <b-icon scale="0.8" icon="question-circle"/></b-dropdown-item>
 
       </b-nav-item-dropdown>
--- a/frontend/src/components/Options.vue
+++ b/frontend/src/components/Options.vue
@@ -3,8 +3,26 @@
     <b-navbar-nav>
       <b-nav-item-dropdown right>
         <template #button-content>
-          <i class="fa fa-cog"></i>
+          <i class="fa fa-bars"></i>
         </template>
+
+        <b-dropdown-group>
+
+        <b-dropdown-header>Links</b-dropdown-header>
+
+        <b-dropdown-item @click="claimSkill(ClaimStage.WaxBridge)">Claim Skill </b-dropdown-item>
+
+        <b-dropdown-item @click.native="$router.push('leaderboard')" class="gtag-link-others" tagname="leaderboard_screen">Leaderboard </b-dropdown-item>
+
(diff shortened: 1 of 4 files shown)


frontend/src/components/Options.vue
```diff
@@ -16,6 +16,8 @@
 
         <b-dropdown-item @click="claimSkill(ClaimStage.WaxBridge)">Claim Skill </b-dropdown-item>
 
+        <b-dropdown-item @click.native="$router.push('leaderboard')">Leaderboard </b-dropdown-item>
```
seiyria, line 19: this will need the google analytics stuff - `gtag-link-others` and `tagname` (I think that's all it needs to be picked up).

frontend/src/components/smart/SkillBalanceDisplay.vue
```diff
@@ -6,8 +6,12 @@
     </div>
 
     <div class="balance-container">
-      <strong class="mr-2 balance-text">Balance</strong>
-      <span class="balance">{{ formattedSkillBalance }}</span>
+      <strong class="mr-2 balance-text">Total Balance</strong>
+      <span class="balance"
+        v-tooltip="{ content: totalSkillTooltipHtml , trigger: (isMobile() ? 'click' : 'hover') }"
+        @mouseover="hover = !isMobile() || true"
+        @mouseleave="hover = !isMobile() || false"
```
seiyria, line 13: || false is redundant

seiyria:
I think that the IGO skill should probably have an icon next to SKILL (like the WAX BNB indicator) that exists and is lit up only when the balance > 0.

seiyria (changes requested):
needs some code changes

poshdan:
I support combining the funds like that, may even make sense to account for the reward pool, since it can be spent from.
Problem is, the ingame only funds can not be used on the market directly, so it would be confusing there.

I wouldn't like the leaderboards being tucked away. At the moment it's very much trivia but I envisioned it having a larger role of showing players the earning potential, where they can see the top earning fights that day or week, etc. Deserves to be easy to find imo.

Would it be possible to have the navbar links break into a new line if it's too long, expanding the height of the header?
Eventually we could condense some of these views where it makes sense, since a lot of them only have a couple of functions.

seiyria:
@poshdan there's an issue up for cleaning up IGO funds for market, so that will eventually clean up

We really don't want a two line navbar, it is super unclean, and we want to combine them where we can into a dropdown. 

Honestly, leaderboards and portal should be hidden because they're far less "useful" than the other views. Leaderboard is not particularly useful, although it is neat, but portal has gotten a shitload of confusing questions because some people don't know what it is, and it's highlighted very prominently.

poshdan:
Well leaderboards would be immensely useful for new users. If only it showed categories that were a "carrot on a stick" for them.
The portal tab is confusing for people because it doesn't explain its purpose (well). In the long run it will be more useful too, because wax won't be the only thing we'll deal with there; it's nice for users to know what's available, without having to stumble onto it through a dropdown. Just my two cents.

frontend/src/components/Options.vue
bilbolPrime: added

frontend/src/components/smart/SkillBalanceDisplay.vue
bilbolPrime: Removed || false from this and from WeaponIcon.vue as well

bilbolPrime:
Perhaps (eventually) instead of a navbar of buttons, each would be a category that's a drop down. Example instead of Plaza it would be Headquarters and when you hover over it it shows Generals (chars now), Blacksmith, Kingdom (assuming the new game will be on same site) etc...
Adventure would show Combat, Raid etc...
And so on
That way the navbar would be compact no matter what is added

bilbolPrime:
> 
> 
> I think that the IGO skill should probably have an icon next to SKILL (like the WAX BNB indicator) that exists and is lit up only when the balance > 0.



> 
> 
> I think that the IGO skill should probably have an icon next to SKILL (like the WAX BNB indicator) that exists and is lit up only when the balance > 0.

Noob issue, I can't find the indicator to mimic it

seiyria:
It's here: https://github.com/CryptoBlades/cryptoblades/blob/main/frontend/src/components/smart/SkillBalanceDisplay.vue#L13

bilbolPrime:
@seiyria 
Updated code as we discussed, please note for moving the portal to the menu:

- Did not add any tracking tags as it didn't have them in the navbar
- Did not add the featureFlagPortal check as I am guessing it's online now

Please let me know if you still have any comments

seiyria (approved):

PR #305, opened 2021-07-18
Add indexes to NFTMarket

brian-man:
This is a proof of concept for a proposal I was discussing about adding indexes to the NFTMarket so that lookups by the common criteria -- individual or in various combinations -- can be done fairly efficiently without a tremendous amount of storage space or time overhead.

The general model is that multi-level mappings allow for specifying zero to five criteria for filtering.  The more filters chosen, the faster fewer the collections that need to be aggregated.  Additionally, pagination limits the amount of aggregation necessary.

The initial commit shows how to add to the index in addListing and how to use it for lookups in getWeaponListingIDsPage which filters on two of the five categories of filtering implemented for weapons.  Any combination of the two filters for characters and five filters for weapons can be used in this model.  For zero filters, it's simpler just to use the already aggregated listedTokenIDs.

### DIFF
--- a/contracts/NFTMarket.sol
+++ b/contracts/NFTMarket.sol
@@ -94,6 +94,19 @@ contract NFTMarket is
     Weapons internal weapons;
     Characters internal characters;
 
+    // Support for 5 categories, each represented by a mapping.
+    // The more selective the filter, the fewer arrays to aggregate.
+    // For characters:
+    //      trait, level, 0, 0, 0
+    // For weapons:
+    //      trait, stars, stat1trait, stat2trait, stat3trait
+    mapping(address =>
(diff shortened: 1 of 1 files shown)


brian-man:
If they really want to go "long" on this market thing (like kroge hinted he wanted) to support all kinds of NFTs, then the approach would be to introduce IIndexable.

That interface would provide the market to query for which bucket an item fits for a given category0, 1, 2, etc.
And then you add category name querying to the interface.

Now you've got a generic mechanism to properly index, filter, and display any NFT.

brian-man:
Private sells don't need to be in the indexes, so would be less stress competing with regular listings and potential save gas when listing a private sale.

Issue #331, opened 2021-07-20
Prevent listings on market of over 10000 skill


seiyria:
No one is going to buy anything for that much.

PR #335, opened 2021-07-20
fix(market): prevent listings over 10000 skill on the client side

kcper:
### All Submissions

https://user-images.githubusercontent.com/9114200/126296280-c3d207ae-6ab5-4b2c-be21-f9c7e855ecf6.mp4

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/331
### Notes

### PR Description

Can't input more than 10000.
Tried a bunch of different ways to prevent inputs > 10000, but only this worked:
`oninput="javascript: if (this.value > 10000) this.value = 10000;"`, looks hacky, but maybe it's how it is in js world.

Note that this will most likely result in conflicts in Market.vue while merging https://github.com/CryptoBlades/cryptoblades/pull/194, which includes most of these changes just with extra target seller field - mainly creating a b-modal template an things around it.

### DIFF
--- a/frontend/src/views/Market.vue
+++ b/frontend/src/views/Market.vue
@@ -274,15 +274,25 @@
                   v-if="activeType === 'weapon'"
                    class="gtag-link-others" tagname="add_listing_weapon"
                   :disabled="selectedNftId === null || selectedNftOnCooldown"
-                  @click="addListingForNft()">List Weapon <b-icon-question-circle :hidden=!weaponMarketTax
+                  @click="showListingSetupModal()">List Weapon <b-icon-question-circle :hidden=!weaponMarketTax
                   v-tooltip.bottom="weaponMarketTax + '% tax (paid by the buyer) will be added to the final price.'"/></b-button>
                 <b-button
                   variant="primary"
                   v-if="activeType === 'character'"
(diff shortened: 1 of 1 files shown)


frontend/src/views/Market.vue
```diff
@@ -274,15 +274,25 @@
...
+                  @click="showListingSetupModal()">List Character <b-icon-question-circle :hidden=!characterMarketTax
                   v-tooltip.bottom="characterMarketTax + '% tax (paid by the buyer) will be added to the final price.'"/></b-button>
+
+                <b-modal class="centered-modal" ref="listing-setup-modal"
+                  @ok="addListingForNft">
+                  <template #modal-title>
+                    Sell {{activeType}}
+                  </template>
+                  <b-form-input type="number"
+                    oninput="javascript: if (this.value > 10000) this.value = 10000;"
```
seiyria, line 293: lets do this a better way. set a `max` value on the form input, and do a `Math.min()` when doing a submit. It'd be doubly good if this was enforced at a contract level, but I'm not sure that will be necessary.

seiyria (changes requested):
the number check has gotta change a bit

seiyria (approved):

PR #353, opened 2021-07-21
fix(character): character name should cut off more gracefully

alistairstpierre:
### All Submissions

image

### Notes

Could potentially update this so that the layout for the first name is centred instead of being off to the side.

### New Feature Submissions

The last name of the character no longer disappears and aesthetically looks a bit better.

### PR Description

Updated CSS so that the last name of the character no longer wraps and disappears.


seiyria:
This'll work for now.

seiyria (approved):

PR #354, opened 2021-07-21
feat(market): show character stamina on market

kcper:
### All Submissions

Market | Plaza:
image image

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/316
### Notes

### PR Description

Stamina shown only in the market character card, as showing by default would duplicate the information we have in the character picker.

Implementation + cleanup of some unused code. 

Stamina calculated from staminaTimestamp which is available in character object we're already using.
If that's not ideal I can rework with fetching stamina for each character.


seiyria (approved):
hell yeah

Issue #349, opened 2021-07-21
Pagination bug


Nyawmeowww:
Pagination visual and functional bug. In the first few seconds of the video , I switched to page 2 of "Earth" and then changed the filter of Element Type to "Fire". The page it fetched is prolly still 1 without resetting the page number (indicated in the video page number is 2) or page is 2 and it doesn't load the other pages.

--it stops working (or takes time to work) when you switch between different combination of filters--

https://user-images.githubusercontent.com/31758395/126416761-6ab070ad-ea39-4ca3-81ab-2ccb30e457a2.mp4

brian-man:
Looks like @pohika did this change.

PR #364, opened 2021-07-21
fix(selectors): pagination should work correctly for all cases

pohika:
#349 reset pagination when change filter, fix filter load incorrectly…

### All Submissions

https://user-images.githubusercontent.com/80157646/126486436-3fdafb04-f212-4466-8635-46727b0d2eb1.mov

### PR Description
- Reset pagination when change filter
- Fix character filters load incorrectly from local when reload market page
- Change fixed filter options to enum style


frontend/src/utils/enum.ts
```diff
@@ -0,0 +1,14 @@
+export const EnumPriceSort = {
```
seiyria, line 1: please rename this: this isn't a real enum. or, make it a real enum with typescript. also, the file should be something more descriptive than `enum.ts`.

seiyria (changes requested):
please fix the above

frontend/src/utils/enum.ts
pohika: Read and think enum not much help project so I reverted it. Thank you

frontend/src/components/smart/CharacterList.vue
```diff
@@ -165,15 +155,24 @@ export default {
...
   },
 
   components: {
     CharacterArt,
   },
 
   mounted() {
-    this.levelFilter = localStorage.getItem('character-levelfilter') || '';
-    this.elementFilter = localStorage.getItem('character-elementfilter') || '';
+    this.levelFilter = sessionStorage.getItem('character-levelfilter') || '';
```
seiyria, line 174: please do not change where things are stored. they should stay in localstorage for now.

frontend/src/components/smart/CharacterList.vue
```diff
@@ -165,15 +155,24 @@ export default {
 
       this.$emit('character-filters-changed');
     },
+
+    priceSortText(type) {
```
seiyria, line 159: this could be done a bit better:

```
const sorts = [
 { name: 'Any', dir: 0 },
 { name: 'Price: Low -> High', dir: 1 },
 { name: 'Price: High -> Low', dir: -1 },
]
```
then use that. then we can add more sorts later without so much boilerplate.

frontend/src/components/smart/CharacterList.vue
seiyria: unless it is sessionstorage everywhere else that is, but I don't think we've done that yet.

frontend/src/components/smart/WeaponGrid.vue
```diff
@@ -272,6 +263,15 @@ export default Vue.extend({
       this.$emit('weapon-filters-changed');
     },
 
+    priceSortText(type: string): string {
```
seiyria, line 266: same as above, use the sorts array instead of this so we don't need this function - I'm not sure why it's here twice.

seiyria (changes requested):

seiyria (approved):

Issue #363, opened 2021-07-21
add a flag to pull market info from the blockchain for local development


seiyria:
should default to off so it doesn't affect prod

bilbolPrime:
Will give this a shot and see what happens.
Looks like an env flag

PR #369, opened 2021-07-21
feat(localdevelopment): can now use blockchain for local dev market testing

bilbolPrime:
### All Submissions

No UI changes

### New Feature Submissions

* [ ] #363 

### Notes

### PR Description

Added a flag VUE_APP_FEATURE_FLAG_MARKET_USE_BLOCKCHAIN that, when set to true, switches to market to local chain to allow devs to work with market.
Tried to keep it as clean as possible but some ugly ifs and elses were inevitable.

**Could not test API locally, please allow a good buffer on testnet before going to prod so that no bugs arise on prod.**


frontend/src/views/Market.vue
```diff
@@ -675,6 +676,21 @@ export default Vue.extend({
       this.marketOutcome = null;
       this.waitingMarketOutcome = true;
 
+      if(useBlockchain === true)
```
seiyria, line 679: `=== true` is superfluous in this case

seiyria: always use braces, no exceptions

frontend/src/views/Market.vue
```diff
@@ -762,17 +813,26 @@ export default Vue.extend({
       this.waitingMarketOutcome = true;
 
       try {
-        const results = this.activeType === 'weapon' ?
-          await this.searchWeaponListingsBySeller(this.search):
-          await this.searchCharacterListingsBySeller(this.search);
-        this.searchResults = results;
+        if(useBlockchain === true){
```
seiyria, line 816: `=== true is superfluous`

frontend/src/views/Market.vue
```diff
@@ -787,13 +847,20 @@ export default Vue.extend({
 
       this.waitingMarketOutcome = true;
 
-      const results = this.activeType === 'weapon' ?
-        await this.searchWeaponListingsBySeller(this.defaultAccount) :
-        await this.searchCharacterListingsBySeller(this.defaultAccount);
+      if(useBlockchain === true){
```
seiyria, line 850: `=== true`

frontend/src/views/Market.vue
seiyria: split this into two functions

frontend/src/views/Market.vue
seiyria: split this into two functions

seiyria (changes requested):
I think instead of having blockchain checks in the relevant functions, I'd split that function into two, and call the respective function. This is done in a few places but not everywhere.

frontend/src/views/Market.vue
bilbolPrime: done

frontend/src/views/Market.vue
bilbolPrime: done

frontend/src/views/Market.vue
bilbolPrime: done

seiyria (approved):

Issue #373, opened 2021-07-21
PVE004: improved loginc in removerewardweapon


seiyria:
![image](https://user-images.githubusercontent.com/763609/126534254-19dec50e-9d84-4ae8-9d81-1b60c50547c2.png)

raymond-h:
I think the actually appropriate thing would've been to use `EnumerableSet.UintSet` from OpenZeppelin to store the weapon drops, but the suggested fix suffices for now.

PR #379, opened 2021-07-21
PVE004: improved loginc in removerewardweapon

kcper:
### All Submissions

N/A
### New Feature Submissions

#373 
### Notes

### PR Description

Copied last element into index that is getting removed and removed last element.

### DIFF
--- a/contracts/raidBasic.sol
+++ b/contracts/raidBasic.sol
@@ -103,7 +103,7 @@ contract RaidBasic is Initializable, Raid {
         require(index < weaponDrops.length, "Index out of bounds");
 
         weaponDrops[index] = weaponDrops[weaponDrops.length - 1];
-        delete weaponDrops[weaponDrops.length - 1];
+        weaponDrops.pop();
     }
 
     function getWeaponDrops() public view returns(uint256[] memory) {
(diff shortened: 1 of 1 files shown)


contracts/raidBasic.sol
```diff
@@ -102,7 +102,8 @@ contract RaidBasic is Initializable, Raid {
     function removeRewardWeapon(uint256 index) public restricted {
         require(index < weaponDrops.length, "Index out of bounds");
 
-        delete weaponDrops[index];
+        weaponDrops[index] = weaponDrops[weaponDrops.length - 1];
+        delete weaponDrops[weaponDrops.length - 1];
```
raymond-h, line 106: The logic from before this PR and after this PR are equivalent so as far as the PVE is concerned, this PR is good, but it's actually probably not quite the logic what we wanted in the first place. The length of the array is left unchanged and there's just a zeroed-out spot in the array instead.

We should call `weaponDrops.pop()` after copying the last element - see suggestion. Popping the last element implicitly calls `delete` on the last element as well.

```suggestion
        weaponDrops[index] = weaponDrops[weaponDrops.length - 1];
        weaponDrops.pop();
```

raymond-h (changes requested):

contracts/raidBasic.sol
kcper: Implemented your suggestion, thanks.

raymond-h (approved):

PR #389, opened 2021-07-21
PVE008: Move characterLimit member from cryptoblades.sol to characters.sol

brian-man:


brian-man:
@raymond-h Do we need to leave (but marked deprecated or something) the characterLimit config member in cryptoblades.sol?

contracts/cryptoblades.sol
```diff
@@ -85,7 +84,6 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
     }
 
     // config vars
-    uint characterLimit;
```
raymond-h, line 88: State variables can never be removed due to proxy contract upgradeability. Instead, add `// UNUSED; KEPT FOR UPGRADEABILITY PROXY COMPATIBILITY` as a note above it.

contracts/cryptoblades.sol
```diff
@@ -388,8 +386,8 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
     }
 
     function mintCharacter() public doesNotHaveMoreThanMaxCharacters oncePerBlock(msg.sender) requestPayFromPlayer(mintCharacterFee) {
-        require(characters.balanceOf(msg.sender) < characterLimit,
-            string(abi.encodePacked("You can only have ",characterLimit," characters!")));
+        require(characters.balanceOf(msg.sender) < characters.characterLimit,
+            string(abi.encodePacked("You can only have ",characters.characterLimit," characters!")));
```
raymond-h, line 390: I believe state variables of other contracts cannot be accessed like this. They need to be treated like getter functions. Also requires the state variable in question to be `public`.

```suggestion
        require(characters.balanceOf(msg.sender) < characters.characterLimit(),
            string(abi.encodePacked("You can only have ",characters.characterLimit()," characters!")));
```

contracts/cryptoblades.sol
```diff
@@ -452,7 +450,7 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
     }
 
     modifier doesNotHaveMoreThanMaxCharacters() {
-        require(characters.balanceOf(msg.sender) <= characterLimit, "Too many characters owned");
+        require(characters.balanceOf(msg.sender) <= characters.characterLimit, "Too many characters owned");
```
raymond-h, line 453: I believe state variables of other contracts cannot be accessed like this. They need to be treated like getter functions. Also requires the state variable in question to be `public`.

```suggestion
        require(characters.balanceOf(msg.sender) <= characters.characterLimit(), "Too many characters owned");
```

contracts/characters.sol
```diff
@@ -95,6 +97,8 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
 
     Promos public promos;
 
+    uint characterLimit;
```
raymond-h, line 100: `uint` is shorthand for `uint256`, and I realize the original code in `cryptoblades.sol` itself used `uint` for this, but I think we should prefer explicitly stating it.

```suggestion
    uint256 characterLimit;
```

contracts/characters.sol
```diff
@@ -24,6 +24,8 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
         __AccessControl_init_unchained();
 
         _setupRole(DEFAULT_ADMIN_ROLE, msg.sender);
+
+        characterLimit = 4;
```
raymond-h, line 28: For now I don't intend to enforce it, but for the future: initialization of new state variables have to be done by a brand-new function as the initializer function was called exactly once - during the first deploy - and will never be called again.

Our convention has been to add a `migrateTo_XXXXXXX` function that gets called during deployment of the new version of the contract, where `XXXXXXX` is the 7 first characters of some relevant Git commit hash that introduced the changes. You can find examples by looking in several other files.

Do feel free to implement that if you wish, but once again, I don't intend to enforce it for now.

raymond-h (changes requested):

raymond-h (approved):

PR #396, opened 2021-07-22
PVE001: Add minimumStakeAmount to staking contracts

brian-man:


contracts/staking/StakingRewardsUpgradeable.sol
```diff
@@ -64,6 +66,7 @@ contract StakingRewardsUpgradeable is
         rewardsToken = IERC20(_rewardsToken);
         stakingToken = IERC20(_stakingToken);
         rewardsDistribution = _rewardsDistribution;
+        minimumStakeAmount = _minimumStakeAmount;
```
raymond-h, line 69: As a note, see https://github.com/CryptoBlades/cryptoblades/pull/389#discussion_r675188791. Again, won't actually enforce it for now, so it's just something to keep in mind for the future.

contracts/staking/StakingRewardsUpgradeable.sol
```diff
@@ -31,6 +31,7 @@ contract StakingRewardsUpgradeable is
     uint256 public periodFinish;
     uint256 public override rewardRate;
     uint256 public override rewardsDuration;
+    uint256 public override minimumStakeAmount;
```
raymond-h, line 34: New state variables must be added after any and all existing state variables, for proxy contract upgradeability reasons,

raymond-h (changes requested):

raymond-h:
It is entirely possible that you already know, and I recognize that in this particular case (since I have stated I was not gonna enforce the migrateTo-function-instead-of-modifying-initializer stuff for now) it was necessary to do it, but I'll mention it just for posterity's sake: if it isn't necessary, do not modify existing migration files. Create a new one instead if necessary.

For now though, I'll revert the migration files myself when I prep this PR to be deployed (which includes the addition of a new `migrateTo` function as well), so you needn't worry about it.

raymond-h (approved):

Issue #386, opened 2021-07-21
Character search filter is not resetting


crypto-smoke:
Character level doesnt change in the api request when changing it on the site

![image](https://user-images.githubusercontent.com/86024507/126562919-824f37a8-3ded-4102-8161-7641a1592ebb.png)

PR #399, opened 2021-07-22
#386 fix character filter not working

pohika:
### PR Description
fix character filter not working because filter not save in sessionStorage correctly

### DIFF
--- a/frontend/src/components/smart/CharacterList.vue
+++ b/frontend/src/components/smart/CharacterList.vue
@@ -144,8 +144,8 @@ export default {
     getCharacterArt,
 
     saveFilters() {
-      localStorage.setItem('character-levelfilter', this.levelFilter);
-      localStorage.setItem('character-elementfilter', this.elementFilter);
+      sessionStorage.setItem('character-levelfilter', this.levelFilter);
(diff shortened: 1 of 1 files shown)


frontend/src/components/smart/CharacterList.vue
```diff
@@ -169,8 +169,8 @@ export default {
   },
 
   mounted() {
-    this.levelFilter = sessionStorage.getItem('character-levelfilter') || '';
-    this.elementFilter = sessionStorage.getItem('character-elementfilter') || '';
+    this.levelFilter = localStorage.getItem('character-levelfilter') || '';
+    this.elementFilter = localStorage.getItem('character-elementfilter') || '';
     if(this.isMarket) {
```
pohika, line 174: revert code to this commit: 3fc33b4bcea38ed0bc37a9311d18c3fabac1ff49

seiyria: we want to be using the same kind of storage everywhere. if we're using session storage, we should continue to do so.

pohika: I know, this is mistake I made change in this commit: 26907d3fe01922d7f9ebf8eb5056803a0af9be4e
So just recover to original code

PR #405, opened 2021-07-22
Update CombatResults.vue

alistairstpierre:
### All Submissions

fight

### PR Description

Fix text overlapping on mobile when winning a fight, still works on other devices like desktop.


kcper:
Cool, it also works when you win a fight and there's more text to it like skill and exp earnings, right?

alistairstpierre:
Yup! Tested that too, works fine there.

seiyria (approved):

PR #409, opened 2021-07-22
feat(market): show sold banners over items that are sold but not removed from the DB yet

### LINKED ISSUE #332: If item price is 0, remove or otherwise mark it as sold
Due to the delay between the database being updated and getting to the user, you will see items on the market with a price of 0. This price is fetched from the contract, which means that item was sold or delisted. If an item price is 0 we should not display it or put an overlay on it marking it sold. Maybe a big red X with SOLD or something.

viruspowaa:
### All Submissions

image

### New Feature Submissions

fix #332 for character and weapon

### PR Description

Add "SOLD" banner for characters AND weapons already sold in market
Manage with "purchase" button


frontend/src/views/Market.vue
```diff
@@ -83,12 +83,17 @@
...
+                      <span v-if="convertWeiToSkill(nftPricesById[id]) !== '0'">
+                        {{ convertWeiToSkill(nftPricesById[id]) | maxDecimals(2) }} SKILL
+                      </span>
                     </span>
                     <span class="d-block text-center" v-else>Loading price...</span>
                   </div>
                 </template>
 
+                <template #sold="{ character: { id } }">
+                  <div class="sold" v-if="nftPricesById[id] && convertWeiToSkill(nftPricesById[id]) === '0'"><span>SOLD</span></div>
```
seiyria, line 95: can you do text-transform: uppercase instead of uppercasing the word?

viruspowaa: Of course you'r the boss :)

seiyria (approved):

Issue #282, opened 2021-07-14
Using purchaseListing with unclaimed Skill doesn't work


brian-man:
Repro steps:
1. Create account
2. Create character
3. Fight until you have some unclaimed Skill
4. Remove all Skill tokens from account
5. Use market to purchase an NFT for <- your unclaimed Skill balance
PROBLEM 1: Metamask prompts for Skill spend approval.
After accepting that (for no good reason)...
PROBLEM 2: Metamask prompts for 0.285 BNB (the "fail" amount) and then fails even if you accept that.

defistar:
@brian-man 

while reproducing steps:

**Step-5:**

Use market to purchase an NFT for <- your unclaimed Skill balance
PROBLEM 1: Metamask prompts for Skill spend approval.

**Unclaimed skill balance (Spending approval) is being done from :**

`await SkillToken.methods
          .approve(NFTMarket.options.address, maxPrice)
          .send(defaultCallOptions(state));`

[approve goes thru even if the SKILL balance in metamask account is 0]

**PROBLEM 2:** Metamask prompts for 0.285 BNB (the "fail" amount) and then fails even if you accept that.

This fails as the SKILL tokens are not yet claimed & Account from which Buy is triggered has SKILL balance 0.
In smart-contract: `NFTMarket.sol`

**function:**  `purchaseListing` does attempt to transfer SKILL token from wallet-account to seller &b tax-recipient

      `
        skillToken.transferFrom(msg.sender, taxRecipient, taxAmount);
        skillToken.transferFrom(
            msg.sender,
            listing.seller,
            finalPrice.sub(taxAmount)
        );
`

- This is where it gets reverted and throw Error:

`Revert reason: ERC20: transfer amount exceeds balance`

----------------------------------------------------------------------------------------

@seiyria 

- what should be the expected behaviour?

- should there be an attempt to use unclaimed SKILL tokens to buy the NFT? (in case the Wallet-Account has 0 tokens )?

- should there be a check for available SKILL balance of the msg.sender / Wallet Account?

** My apologies for delay in debugging as this is first time, that i am exploring the UserInterface and i didn't knew that i need to List the NFT before making attempt to buy in Market.

Nizzypal:
How does one get banned from the marketplace?
![image](https://user-images.githubusercontent.com/2307123/127798609-2d16bc7e-04a4-42a4-8c27-a1e7a7e08efd.png)

seiyria:
[...]

PR #438, opened 2021-07-23
fix: issue with purchaseListing function using unclaimed SKILL

magicianlee007:
#282

### DIFF
--- a/contracts/NFTMarket.sol
+++ b/contracts/NFTMarket.sol
@@ -10,6 +10,7 @@ import "../node_modules/@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "../node_modules/@openzeppelin/contracts/token/ERC721/IERC721.sol";
 import "../node_modules/abdk-libraries-solidity/ABDKMath64x64.sol";
 import "./interfaces/IPriceOracle.sol";
+import "./interfaces/ICryptoBlades.sol";
 import "./characters.sol";
 import "./weapons.sol";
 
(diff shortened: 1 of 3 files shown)


contracts/NFTMarket.sol
```diff
@@ -488,12 +489,46 @@ contract NFTMarket is
...
+
+            uint256 taxAmountUserWallet = taxAmount.sub(taxAmountInGameOnlyFunds).sub(taxAmountTokenRewards);
+
+            skillToken.transferFrom(msg.sender, taxRecipient, taxAmountUserWallet);
+            skillToken.transferFrom(
+                msg.sender,
+                listing.seller,
+                fromUserWallet.sub(taxAmountUserWallet)
+            );
+        }
```
raymond-h, line 531: I believe this code could be greatly simplified such that the near exact same code isn't copy-pasted several times over.

contracts/NFTMarket.sol
```diff
@@ -488,12 +489,46 @@ contract NFTMarket is
...
+        } else {
+            uint256 taxAmountInGameOnlyFunds = fromInGameOnlyFunds.mul(taxAmount).div(finalPrice);
+            ICryptoBlades(taxRecipient).transferInGameOnlyFunds(msg.sender, taxRecipient, taxAmountInGameOnlyFunds);
+            ICryptoBlades(taxRecipient).transferInGameOnlyFunds(msg.sender, listing.seller, fromInGameOnlyFunds.sub(taxAmountInGameOnlyFunds));
+
+            uint256 taxAmountTokenRewards = fromTokenRewards.mul(taxAmount).div(finalPrice);
+            ICryptoBlades(taxRecipient).transferTokenRewards(msg.sender, taxRecipient, taxAmountTokenRewards);
+            ICryptoBlades(taxRecipient).transferTokenRewards(msg.sender, listing.seller, fromTokenRewards.sub(taxAmountTokenRewards));
+
+            uint256 taxAmountUserWallet = taxAmount.sub(taxAmountInGameOnlyFunds).sub(taxAmountTokenRewards);
```
raymond-h, line 523: Cast `taxRecipient` just once and store the result in a local variable.

Also, we do not want to give token rewards or IGO to `taxRecipient` since that's gonna be the main Cryptoblades contract.

contracts/interfaces/ICryptoBlades.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity ^0.6.5;
+
+interface ICryptoBlades {
```
raymond-h, line 3: Make the CryptoBlades contract properly implement the interface, and mark functions as overridden as needed by the compiler.

raymond-h (changes requested):

kcper:
Outdated.

Issue #430, opened 2021-07-23
Market buyer price precision fix


bilbolPrime:
When listing for less than 0.01, buyer price shows as 0.00
![image](https://user-images.githubusercontent.com/25177987/126821124-a40a80ca-9c00-4361-8db2-08cbdc3c3350.png)

Should show exact buyer price.

_This is me cleaning my mess, no need for bounties on this one_

PR #445, opened 2021-07-24
fix(market): sell popup precision was off

bilbolPrime:
### All Submissions

* [ ] 
image
image
image
image
image
image

### New Feature Submissions

* [x] #430 

### Notes

### PR Description

We added support on market to properly display < 0.01 SKILL items so I figured we need to do it here as well.
javascript just doesn't like dividing by 100 or multiplying by 0.01 and ends up giving like 0.12100000000000001. The simplest solution I could come up with because I don't think this is worth adding some custom libraries or whatever was to just toFixed(8) and trim any trailing 0s. Went with 8 because that's what we show in the tool tip in market for the _exact_ value.

_Me cleaning my mess, no bounties on this one_

### DIFF
--- a/frontend/src/views/Market.vue
+++ b/frontend/src/views/Market.vue
@@ -1042,7 +1042,7 @@ export default Vue.extend({
     },
 
     calculatedBuyerCost(listedPrice: number): string {
-      return (0.01 * listedPrice * (100 + parseFloat(this.activeListingMarketTax()))).toFixed(2);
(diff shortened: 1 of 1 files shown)


seiyria (approved):
regex is a wild choice here, why not substring?

bilbolPrime:
> 
> 
> regex is a wild choice here, why not substring?

Wanted to trim all trailing zeroes and couldn't find a trim end for string in javascript.
Only solution I could find online (aside from math libraries) is to use regex.

If this can be accomplished with substring I can open another PR

seiyria:
I don't need another pr, was just curious!

Issue #330, opened 2021-07-19
Allow hyperlinking to a specific weapon or character or (future nft)


crypto-smoke:
Have a page that displays a weapon or character card like it is in the game, that can be directly linked to. Users can brag and show off their NFTs.

seiyria:
same implementation as #329

brian-man:
@Sellie96 Is anybody active on this one?

Sellie96:
@brian-man the functionality is still there it was missed off of the new nav ui. Chris has already done the work to add it back to the menu.

PR #447, opened 2021-07-24
feature/330 - Allow hyperlinking to a specific weapon or character or (future nft)

FilipBartos:
### All Submissions

NFT Listing

image

NFT share action

image

NFT copied link result

image

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/330

### Notes

### PR Description

(insert description here)


seiyria:
Can we get the share button to the left of the clear filters button? Having it on every weapon is a bit much. With it being there, it will allow sharing a set of market results better. We can add a share id option on the search nfts page, since that one is look up by id.

frontend/src/router.ts
```diff
@@ -32,7 +32,8 @@ function createRouter() {
   let marketRoutes: RouteConfig[] = [];
   if(featureFlagMarket) {
     marketRoutes = [
-      { path: '/market', name: 'market', component: Market }
+      { path: '/market/:category/:type/:id', name: 'marketSearch', component: Market },
```
seiyria, line 35: this should really all be query params on /market. we have a lot of search criteria to go through.

FilipBartos:
@seiyria sure, what about separate view - something like "seller's profile" that would list characters/weapons from specified seller address? Something like at the screens below:

image
image

seiyria:
Sure, we could do something like that. What you'll want to do is accept query parameters for everything that goes into the search query - seller address, element, minlevel, maxlevel, etc, and use that. then people can share their current settings much more easily than having more market routes.

seiyria (changes requested):
conflicts need to be fixed, as well as the other change I requested

frontend/src/views/Seller.vue
```diff
@@ -0,0 +1,81 @@
...
+  },
+
+  methods: {
+    ...(mapActions(['fetchMarketNftIdsBySeller' ])),
+
+    async fetchData () {
+      const route = this.$route;
+      const sellerAddress = route.params.sellerAddress;
+      this.items = await this.searchListingsBySellerThroughChain(sellerAddress);
+      console.warn(this.items);
```
seiyria, line 68: please remove console statements

ultradeq:
Hi @seiyria , what about placing the "share market link" (or something like "share my market") to the character display area (see attached image)? It would then copy the url .../market/seller/:accountAddress (where the user's items for sale would be listed) to the clipboard.
cb-share-market-link

seiyria:
I think that location would be fine, but again, everything should be a query parameter. There should be no new routes for this.

kcper:
As there were no updates here, I resolved the issue using "..." menu with "Copy link" option and NFT Display component here: https://github.com/CryptoBlades/cryptoblades/pull/851

PR #469, opened 2021-07-27
Remove transfer cooldown from characters

brian-man:


seiyria:
ui-wise this is probably fine but it does need to pass the build

gmigson:
saludos. soy usuario de CRYTOBLADES, 
realice un enlistado en el market de un personaje y ahora no aparece en ningún lado..

seiyria:
@gmigson this issue tracker is EN only. also, it's not for general users.

brian-man:
@KR Please check the contract stuff carefully.  I don't fully grok the proxy upgradability rules.

raymond-h (approved):

PR #473, opened 2021-07-27
Update Combat.vue

alistairstpierre:
### All Submissions

https://media.giphy.com/media/rmmMR9Ht6hJqPYH5De/giphy.gif

### PR Description

Removed interactions with modal and put it on a timer to fix usability issues.

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -5,10 +5,11 @@
         <div class="col error">Error: {{ error }}</div>
       </div>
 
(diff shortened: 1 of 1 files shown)


seiyria:
2000 ms is really not that long - people will not be able to process it quickly enough. Removing external interaction is fine as long as they have to hit the close button, but auto closing it without a way to review it isn't a good idea.

seiyria (changes requested):
2000ms is not enough time

seiyria:
Going to close this for now. I don't think this approach will work out well. Fight history might sort us out better.

PR #474, opened 2021-07-27
fix(combat): shrink combat ui

wintersummermint:
### All Submissions

Too large for smaller viewport / screens Issue : 
Screen Shot 2021-07-27 at 6 18 56 PM

FIX : 

Screen Shot 2021-07-27 at 6 18 42 PM

Mobile View : 
qxTXAQfiJw

Tablet View :  1025
Screen Shot 2021-07-27 at 6 12 31 PM

### New Feature Submissions

### Notes

### PR Description

(insert description here)


seiyria:
this is better than before, but I would argue that twiddling units is a pain, these should be col-md-3 col-sm-6 col-xs-12 probably. I will still merge this but if it continues to be a problem we're going to stick to the bootstrap way of doing things.

seiyria (approved):

PR #494, opened 2021-07-28
Get a free Weapon Forge if you win an Unlikely fight

procart:
Mint a sword if you win an Unlikely fight.


sarimash:
@brian-man any idea how bad this would be on gas costs? This looks like a lot of operations to run on every attack, there might be some shortcuts to take to prevent some of the checks but I haven't had the chance to look at it to make any suggestions.

sarimash:
I missed a lot of the discussion around this, but are we really wanting to mint a weapon to the burn address every fight? I thought the requirement was to mint a weapon on unlikely fights, to player on win and to burn (or pool or whatever) on loss.

procart:
It's around 30k gas

sarimash:
Also, I think we want to make the burn address configurable so Kroge (or sei or whoever) can point it at a giveaway collection pool temporarily and repoint it, obviously the configuration needs to be limited in access similar to MintWeaponWithStars

seiyria:
Correct, we want the burn address to be configurable so we can recover the weapons to use in giveaways or others.  the default address I'd like is 0xc4f7460D6162d8e1Bde3A295B79be8B0dDC8F66c

procart:
Added the configurable burn address

brian-man:
> @brian-man any idea how bad this would be on gas costs? This looks like a lot of operations to run on every attack, there might be some shortcuts to take to prevent some of the checks but I haven't had the chance to look at it to make any suggestions.

I haven't run the numbers but I stated very clearly in dev-round-table channel my objections and some of the reasons for it.

Inflating the cost of the most used contract function -- fight() -- on BSC seems like a horrible idea.  We did a lot of work to get more efficient in that function and we're throwing it away.

Also I don't believe in the fundamentals of the change: it doesn't produce the desired affect, has a bad affect on the contract and the chain, lowers market value of weapons, lowers the market price of items, lowers the market proceeds from weapon listings, etc.  Basically there is a long list of bad effects and no good effects.

Lastly, you're directly rewarding people for choosing the worst combat target.  That's bizarre.

procart:
Is there anything else here that needs to be done?

procart:
Squashed and solved merge conflicts

contracts/cryptoblades.sol
```diff
@@ -271,6 +271,49 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
 
+    mapping (address => uint32) tickets;
+
+    function spendTicket(uint32 num)
+        public
+    {
+        require (tickets[msg.sender] >= num, "You don't have enough tickets");
+
+        for (int i = 0; i < num; i++) {
+            weapons.mint(msg.sender, randoms.getRandomSeed(msg.sender));
```
raymond-h, line 282: This will give every single weapon the exact same seed and thus make them identical. The random seed should be combined with `i` and passed to `keccak256()` to make a unique seed for each weapon.

contracts/cryptoblades.sol
```diff
@@ -271,6 +271,49 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         require(foundMatch, "Target invalid");
     }
 
+    mapping (address => uint32) tickets;
```
raymond-h, line 274: This should be added in a brand-new contract instead of the main game contract. I'll have to get back to you whether I just make those myself, let another PR create them or have this PR create them.

contracts/cryptoblades.sol
```diff
@@ -271,6 +271,49 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
+            rollingTotal = ABDKMath64x64.toInt(temp.mul(ABDKMath64x64.fromUInt(1000)));
+        } else {
+            int128 temp = enemyMin.sub(playerMin).div(playerRange);
+            temp = temp.add(ABDKMath64x64.fromUInt(1).sub(temp).mul(enemyMax.sub(playerMax).div(enemyRange)));
+            temp = temp.add(ABDKMath64x64.fromUInt(1).sub(temp).mul(ABDKMath64x64.fromUInt(50).div(ABDKMath64x64.fromUInt(100))));
+            temp = ABDKMath64x64.fromUInt(1).sub(temp);
+            rollingTotal = ABDKMath64x64.toInt(temp.mul(ABDKMath64x64.fromUInt(1000)));
+        }
+
+        return rollingTotal <= 300 ? true : false;
```
raymond-h, line 314: This logic probably does not require use of `ABDKMath64x64` to implement at all, and removing all uses of it should lead to huge gas savings.

raymond-h (changes requested):

contracts/cryptoblades.sol
procart: Fixed this.

contracts/cryptoblades.sol
procart: It may be possible to convert the calculations, but there won't be big gas benefits. I did a quick test in remix and the gas changed from 28277 to 25490.

raymond-h (approved):

PR #508, opened 2021-07-28
Add contract support for bulk weapon burning

brian-man:


contracts/cryptoblades.sol
```diff
@@ -436,6 +436,15 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         weapons.burn(burnID);
     }
 
+    function burnWeapons(uint256[] memory burnIDs) public
+            doesNotHaveMoreThanMaxCharacters
+            isWeaponsOwner(burnIDs) requestPayFromPlayer(ABDKMath64x64.fromUInt(burnWeaponFee.mulu(burnIDs.length))) {
+        _payContract(msg.sender, ABDKMath64x64.fromUInt(burnWeaponFee.mulu(burnIDs.length)));
```
raymond-h, line 442: I don't think `ABDKMath64x64.fromUInt(burnWeaponFee.mulu(burnIDs.length))` will have the desired effect (for example, `3.5 * 3 = 10.5`, which would get rounded off to `10`). Instead I recommend `burnWeaponFee.mul(ABDKMath64x64.fromUInt(burnIDs.length))`.

```suggestion
            isWeaponsOwner(burnIDs) requestPayFromPlayer(burnWeaponFee.mul(ABDKMath64x64.fromUInt(burnIDs.length))) {
        _payContract(msg.sender, burnWeaponFee.mul(ABDKMath64x64.fromUInt(burnIDs.length)));
```

raymond-h (approved):

raymond-h (changes requested):
See comment above.

contracts/cryptoblades.sol
brian-man: > burnWeaponFee.mul(ABDKMath64x64.fromUInt

Great catch.  Thanks!

raymond-h (approved):

PR #520, opened 2021-07-29
Adjust custom stamina token gain calculation

brian-man:
When calculating the token gain for a custom stamina fight, the gas offset should only be included once.


newb23 (approved):
I concur; good catch!

brian-man:
> I concur; good catch!

Kroge caught it.  I just helped fix it. :)

newb23:
Nevertheless. I am thankful and trapped at work, work. A good catch.

poshdan:
Seems good. Obligatory round on testnet before going on mainnet

PR #522, opened 2021-07-29
Create es.json

DrGabo:
Spanish Language - Working on it with VirusPowaa.


seiyria:
going to merge this. if there are inconsistencies or more work needed we can fix it when the i18n PR moves forward

Issue #535, opened 2021-07-30
Prevent players from selecting a higher stamina cost for a fight than they have available


crypto-smoke:
Some players are selecting 200 stamina cost for a fight despite not having 200 stamina and are wasting gas with failed transactions.

0xroy:
How about this? I remove the available options depending on the Character Stamina
![Issue#535](https://user-images.githubusercontent.com/87876521/127660729-24264a59-264f-4e38-a84b-ef75d3df58e3.png)

crypto-smoke:
that sounds good to me

seiyria:
lgtm

PR #540, opened 2021-07-30
fix(combat): disallow bad stamina values for fight

0xroy:
### All Submissions

### New Feature Submissions

* [ ] Issue #535

### Notes

### PR Description

(insert description here)

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -19,14 +19,7 @@
               <div class="message-box" v-if="currentCharacter && currentCharacterStamina < staminaPerFight">
                 You need {{ staminaPerFight }} stamina to do battle.
                 <h4>Stamina Cost Per Fight</h4>
-                <b-form-select v-model="fightMultiplier" @change="setFightMultiplier()" class="ml-3">
-                  <b-form-select-option :value="null" disabled>Please select Stamina Cost per Fight</b-form-select-option>
-                  <b-form-select-option value="1">40</b-form-select-option>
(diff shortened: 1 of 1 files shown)


frontend/src/views/Combat.vue
```diff
@@ -343,6 +329,54 @@ export default {
     setFightMultiplier() {
       localStorage.setItem('fightMultiplier', this.fightMultiplier.toString());
     },
+
+    setStaminaSelectorValues(){
```
seiyria, line 333: this needs to be done way better, there is too much repetitive code here. I will assist.

0xroy: Alright I was thinking about something like having an array with all options and then just removign the ones who dont pass

seiyria (approved):

frontend/src/views/Combat.vue
seiyria: take a look at my solution :3

0xroy: Nice solution! I was trying something simmilar just the other way arround I guess
But Im still trying to figure out how to avoid the if else wall :P
image

seiyria: there are other ways, like doing a filter on an array, but I was not feelin it

0xroy:
Sorry for the noob question but how do I link the Issue to this and move it to "Done - Bounty Requested"?

seiyria:
Linking an issue is what you did, I just forgot to do the close comment. As for moving it, I will send you an invite to the github org so you can manage your PR states from now on. cheers!

PR #544, opened 2021-07-31
Remove oncePerBlock()-equivalent logic from fightModifierChecks()

brian-man:
The call to `characters.getFightDataAndDrainStamina() in fight() enforces the stamina requirements.

Additionally, nested through _verifyFight() and verifyFight(), the invocation of getTargetsInternal() ensures that the staminaTimestamp impacts the targets when validating the target passed to fight().

Consequently, there is no additional need for oncePerBlock()-equivalent logic in the fight() method.

This change removes this unneeded gas overhead.


poshdan (approved):
This saves like 5200 gas on fights. Possibly a bit more since my local numbers are always a bit lower than bsc mainnet. (may be up to 7600)

Should make about 0.000026 - 0.000038 BNB difference. That's about 1 cent, and allows us to reduce the gas offset. We had 5.48m transactions yesterday, of which probably about 5m were fights. That's $50k (atm ~900 skill worth) that we could avoid paying out each day.

Nice find, approved

PR #545, opened 2021-08-01
Introduce add and change fees to NFTMarket

brian-man:


brian-man:
Elevator pitch:

I believe the market should have an up-front listing fee.  The current implementation only costs the gas fee for the transaction.

People act stupid when something is "free".  Adding even a tiny listing fee will have two effects [a] make people pay for using the market contract storage and [b] make some people reconsider bulk listing stuff that will never sell (or won't for a long time).

poshdan:
I don't think we should charge extra for changing price. A user should have as little friction as possible in their ability to sell their item for a valid price, we want to encourage them to complete the sale, it's not their fault if its value changes due to supply/demand.

brian-man:
> I don't think we should charge extra for changing price. A user should have as little friction as possible in their ability to sell their item for a valid price, we want to encourage them to complete the sale, it's not their fault if its value changes due to supply/demand.

I think we *want* friction to prevent spamming or stressing the market.

Seiyria and others have done a bunch of work to make the market usable without hitting the contract/chain as often.

Every time a listing has a price change it has a cost/work/performance impact to keeping the market data current.

The BNB cost of the transaction is already several times both the addFee and the changeFee, at least with the initial values I set.

seiyria:
A listing fee is fine, but it should probably be a static price. 0.1 skill for example.

poshdan:
I'm not against the listing fee, and I want to reduce the market traffic of course.
We want users to complete their sales as much as we want to reduce the number of listings so i think a frictionless update price will help with that. Charging extra may discourage it. I can't really say which one is better, since we do gain a bit from the updates if we charge for it, and they may decide to cancel the listing to recoup losses, at which point we get the decreased load anyway.

brian-man:
> A listing fee is fine, but it should probably be a static price. 0.1 skill for example.

I considered that but then it doesn't work with variable market price.  For example, a listing fee of $5.71 seems a bit high (0.1 skill at current market price of $57).  So I went with the model we use for other fees -- "cents"-based and adjustable by game admins.

brian-man:
Kroge requested changing setting the initial changeFee to zero.  I think @poshdan agrees.  Updating...

raymond-h (approved):

PR #546, opened 2021-08-01
Add character buy-back to NFTMarket

brian-man:


contracts/NFTMarket.sol
```diff
@@ -507,6 +531,28 @@ contract NFTMarket is
         );
     }
 
+    function buyBackIfPriceIsLow(IERC721 _tokenAddress, uint256 _id, uint256 _price) private returns (bool) {
+        if(address(_tokenAddress) == address(characters)) {
+            uint8 characterLevel = characters.getLevel(_id);
+            int128 buyBackUSDValue = characterBuyBackNew + characterBuyBackPerLevel.mul(ABDKMath64x64.fromUInt(characterLevel));
+            uint256 buyBackTokenValue = usdToSkill(buyBackUSDValue);
+            if(buyBackTokenValue >= _price) {
+                _tokenAddress.safeTransferFrom(msg.sender, address(0), _id);
```
raymond-h, line 540: The `.safeTransferFrom` function call won't allow passing the zero address as the destination. This will instead require the use of the `_burn` internal function in the NFT contract.

raymond-h (changes requested):

contracts/NFTMarket.sol
```diff
@@ -450,6 +469,11 @@ contract NFTMarket is
         isListed(_tokenAddress, _id)
         isSeller(_tokenAddress, _id)
     {
+        if(buyBackIfPriceIsLow(_tokenAddress, _id, _newPrice)) {
+            cancelListing(_tokenAddress, _id);
```
raymond-h, line 473: Based on the code in both `buyBackIfPriceIsLow` (which burns the chara) and `cancelListing` (which tries to transfer the chara), I have a feeling the chara is always going to be burned before the cancellation tries transferring it, meaning it'll always error out in this case.

contracts/characters.sol
```diff
@@ -140,6 +140,10 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
         return success ? secondsLeft : 0;
     }
 
+    function burn(uint256 tokenId) public restricted {
```
raymond-h, line 143: AFAIK, character-burning may be something ordinary users can do too in the future? I think this function maybe should be renamed to clarify it's for more admin-y purposes. I would like it to also have a dedicated, more specific, role instead of `GAME_ADMIN` that the `restricted` modifier enforces, and maybe also check that `msg.sender` is the owner as well.

The use of `GAME_ADMIN` for everything and anything has been pretty rampant through the entire contract codebase and I'd like to change that.

raymond-h (changes requested):

Issue #548, opened 2021-08-01
Validating target power should hide/disable fight buttons


bilbolPrime:
Before the request to issue a fight is done, front end checks if the targets are valid.

User with slow enough internet can click the button more than once (Easy to simulate with google chrome, set network throttle to slow 3G). This leads to multiple meta mask transaction requests. The user probably confirms the first one then is confused, or confirms multiple, gets an exception and probably thinks they lost stamina.

![image](https://user-images.githubusercontent.com/25177987/127762678-980bf78a-06b2-401f-bab2-65ab9d70409a.png)

The fight buttons should be hidden while the validation is being made to avoid this.

On main branch the buttons get disabled not hidden now but the issue is still their. Reproduced locally by simulating delay with this
![image](https://user-images.githubusercontent.com/25177987/127767424-9efd5197-6861-4d1f-bb8d-0be8f15cdcb5.png)

The waiting for results needs to be set true before the await for validation and set to false if not valid
![image](https://user-images.githubusercontent.com/25177987/127767430-acb8bd9f-7226-417d-a20d-ee2c06151919.png)

PR #551, opened 2021-08-01
fix(combat): when validating targets, disable fight button

bilbolPrime:
### All Submissions

### New Feature Submissions

* [x] #548 

### Notes

### PR Description

waitingResults should be set to true before validating targets to disable buttons. Otherwise players with poor internet can click fight button multiple times and end up issuing multiple transactions leading to exceptions and confusion.

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -278,16 +278,18 @@ export default {
         return;
       }
 
+      this.waitingResults = true;
+
       // Force a quick refresh of targets
       await this.fetchTargets({ characterId: this.currentCharacterId, weaponId: this.selectedWeaponId });
       // If the targets list no longer contains the chosen target, return so a new target can be chosen
       if (!this.targets.find((target) => target.original === targetToFight.original)) {
+        this.waitingResults = false;
         return;
       }
 
       this.fightResults = null;
       this.error = null;
-      this.waitingResults = true;
       this.setIsInCombat(this.waitingResults);
 
       try {


frontend/src/views/Combat.vue
```diff
@@ -278,16 +278,18 @@ export default {
         return;
       }
 
+      this.waitingResults = true;
+
       // Force a quick refresh of targets
       await this.fetchTargets({ characterId: this.currentCharacterId, weaponId: this.selectedWeaponId });
       // If the targets list no longer contains the chosen target, return so a new target can be chosen
       if (!this.targets.find((target) => target.original === targetToFight.original)) {
+        this.waitingResults = false;
```
seiyria, line 287: this still needs to be set outside the return too

bilbolPrime: I think that would enable the buttons again before battle finishes. I never checked what sets it back to false if you battle but it probably has to do with getting fight results back. Not home right now, will check and update when I can.

seiyria: @bilbolPrime let me know the fruits of your research when you can!

bilbolPrime: Hello!
Sorry for delay, hectic time at work
So the reason I never gave this much attention when I did that change was that the method never set it back to false (after await) so I just figured that some callback set it to false. And it's being done in this method:
`async updateResults([fightResults, error]) {
      this.resultsAvailable = fightResults !== null;
      this.waitingResults = fightResults === null && error === null;
      this.setIsInCombat(this.waitingResults);
      if (this.resultsAvailable) this.$bvModal.show('fightResultsModal');
    }`

When the battle is over or an error is thrown that method sets waitingResults to false again. If there is an issue in relying on that method, then it's on prod right now even without this change. updateResults is being watched so I am guessing it's being called whenever fightResults or error changes. The change should not introduce any issue that doesn't already exist and I don't think there is.

seiyria: Sounds good to me, thanks for the analysis!

seiyria (approved):

PR #552, opened 2021-08-01
use target index instead of target power+trait as a fight parameter to save gas

poshdan:
We can use target index instead of packed target value in fights to save ~8.5k gas.

That's about 0.000042 BNB or currently ~1.4 cents.

This way we only calculate the one target instead of all 4 and we skip the verification comparison too.
Players on mainnet will experience reverted fights until both the contract and frontend is in sync. I think that's fine, since they can just not send the transaction when metamask warns them.


raymond-h (approved):

brian-man:
@poshdan The existing app protects against "doing the wrong target" when the hour switches.  Is there a way to provide that with the new model?

PR #557, opened 2021-08-02
fix(combat): can no longer fight when weapon would not be able to fight based on stamina multiplier

KennyUTN:
### Description

In views/Combat.vue the setStaminaSelectorValues() method checks only if the fighter can do the battle, so if you have a sword with 1 durability and a fighter with 160, users are able to start a transaction that will later fail (by choosing 40 stamina, 1 durability sword, and then changing the stamina to a greater number). This will potentially make a lot of users lose unnecessary gas, and it will happen more often with the PVP and Raids because of how they work with stamina.

### Screenshots
I've greyed out the fight button when such a thing happens, so now, if your sword does not have the durability, you cannot start a transaction.

image

### Notes
The "This weapon does not have durability" message still shows, but players were still able to click on "Fight" and start the transaction.

As a suggestion, I would change the max stamina of the fighters to 50, and make every fight cost 10 (also adjust the sword's durability to match). This makes it so players can know how many fights they can do by looking at the tens of their current stamina instead of doing multiplication.

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -108,7 +108,7 @@
                     <big-button
                       class="encounter-button btn-styled"
                       :mainText="`Fight!`"
-                      :disabled="(timeMinutes === 59 && timeSeconds >= 30) || waitingResults"
+                      :disabled="(timeMinutes === 59 && timeSeconds >= 30) || waitingResults || !weaponHasDurability(selectedWeaponId)"
                       @click="onClickEncounter(e)"
                     />
                     <p v-if="isLoadingTargets">Loading...</p>


seiyria:
Wasn't this fixed a while back? Can you check against main to see if this is still a problem?

seiyria:
Specifically in #540

KennyUTN:
It has not been fixed!. It's a similar problem, it can still happen. You need to have a weapon with (1 2 3 4) durability, and a fighter with greater stamina than the weapon. If you first choose a possible fight multiplier for both the sword and the fighter, you can then change the stamina to an amount that the sword cant do and still try to Fight.

seiyria:
Cool, sounds good to me. I notice there is no linked issue, is this a fresh thing we haven't logged yet, or is there an issue I missed?

KennyUTN:
No, there's not. I tried to DM you on discord but then just did the PR!

seiyria:
Thanks. If you're not in our round table chat, please ping me in off topic. I don't take Discord DMs.

seiyria (approved):

KennyUTN:
> Thanks. If you're not in our round table chat, please ping me in off topic. I don't take Discord DMs.

It's the "we share a different discord" so you receive my notifications thing. I'm on the round table already

PR #566, opened 2021-08-02
Remove base weapon power and adjust the per point values

brian-man:


brian-man:
There are many things being discussed in the "Motivating weapon upgrades" thread on Discord.

On specific idea was raised by Shyrus to remove the base weapon multiplier of 1.  This would make star count, stat alignment, and reforging more valuable than they are today.  It would also diminish significantly the effectiveness of 1star and 2star weapons, which is also good.

seiyria (approved):

raymond-h (approved):

contracts/weapons.sol
```diff
@@ -642,6 +653,12 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
         }
     }
 
+    function setBaseWeaponMultiplier(int128 multiplier) public restricted {
+        require(multiplier >= 0, "BaseWeaponMultiplier too low");
```
poshdan, line 657: These require statements will always block the function.
int128 variables aren't inherently handled as fixed64x64 numbers so they can't be used for a fractional comparison
I also don't think these sanity checks should be kept at all.
I'd rather the function parameter took in a decimal number that we then divide by like 10k using ABDK to get the fraction we need.

poshdan (changes requested):
I'm not sure we should add the base multiplier variable if we plan on it being 0.
The fact that this code is anticipating its change, makes me feel like it hasn't been thoroughly checked and finalized.
I'll try to make some time to test this myself but for the time being I'd like to hold it until at least others can take another look.

PR #576, opened 2021-08-03
Draft: Possible fix for mint attack

procart:
Update a second seed in every fight, and use it for random number generation.

### DIFF
--- a/contracts/ChainlinkRandoms.sol
+++ b/contracts/ChainlinkRandoms.sol
@@ -48,6 +48,10 @@ contract ChainlinkRandoms is IRandoms, Pausable, AccessControl, VRFConsumerBase 
         return uint256(keccak256(abi.encodePacked(user, seed, blockhash(block.number - 1))));
     }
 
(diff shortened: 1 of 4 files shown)


poshdan:
I think it would make sense to do the thing entirely in the randoms contract
using the same function, no need for a new one, but adding a restricted modifier, granting only game contracts access and doing
seed = uint256(msg.sender) ^ seed;
plus whatever else to add additional layers (blockhash etc)

procart:
> I think it would make sense to do the thing entirely in the randoms contract
using the same function, no need for a new one, but adding a restricted modifier, granting only game contracts access and doing
seed = uint256(msg.sender) ^ seed;
which would require changing it from a view into a regular function but whatever.
plus whatever else to add additional layers (blockhash etc) 

Yes, moving everything to randoms will be more elegant. I'm not very familiar with the process to grant access to contracts, so feel free to make any changes.

DEntwickler:
Drain much gas cost if token not enough as hot fix. Everything should be fine.

Issue #562, opened 2021-08-02
choose weapon element on forge


seiyria:
blocked by #538 and #529

this should allow you to, for a markup of 100% (total of 2x), choose the element you mint with. the new flow is:

1. hit forge button
2. see popup of "random" (with the cost below it, current price), then each of the elements (with the cost below them, 2x current price)
3. choose forge element or random, then hit a button to confirm

this will need to wait for the 10x forge to be merged because it will need to be factored in there as well.

0xroy:
Maybe something like this with a hover over the lements at the beginning and after selecting one it stays trandformed? 
Also I was thinking maybe have a questionmark png for the random or some more fancy font.
![image](https://user-images.githubusercontent.com/87876521/127998507-48f03d2c-b2b3-454c-92c2-353a81edad5b.png)

seiyria:
I would use the question mark icon.

However if you were to do this, the choices should be one row, not two. Something like

Random fire water thunder earth
X skill | y skill

Forge

PR #582, opened 2021-08-03
feature choose weapon element on forge

0xroy:
### All Submissions

image
image
image

### New Feature Submissions

* [ ] Issue #562

### Notes

### PR Description

(I believe everything needed is done except paying the correct skill amount for minting the weapons if a element was choosen. 
I might be able to get into this tomorrow/ in the next few days but wouldn't mind if someone else wants to do it.)


contracts/cryptoblades.sol
```diff
@@ -209,7 +209,7 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
             // isCharacterOwner(char)
             // isWeaponOwner(wep) {
         fightModifierChecks(char, wep) {
-        require(fightMultiplier >= 1 && fightMultiplier <= 5);
+        require(fightMultiplier > 0);
```
seiyria, line 212: do not revert this change

contracts/weapons.sol
```diff
@@ -236,9 +236,15 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
         return tokenID;
     }
 
-    function getRandomProperties(uint256 stars, uint256 seed) public pure returns (uint16) {
+    function getRandomProperties(uint256 stars, uint256 seed, uint256 choosenElement) public pure returns (uint16) {
+        uint256 trait;
+        if (choosenElement == 4) {
```
seiyria, line 241: "chosen" also we really should not use 4 as the sentinel value for this; it should be higher, like 100. We don't plan to add more elements, but if we did, we don't want to have to redo this as well.

frontend/src/App.vue
```diff
@@ -311,7 +310,7 @@ export default {
     },
 
     async checkNotifications() {
-      const response = await fetch(apiUrl('static/notifications'));
+      const response = await fetch('https://api.cryptoblades.io/static/notifications');
```
seiyria, line 313: this was intentional, please revert it

frontend/src/store.ts
```diff
@@ -953,7 +953,7 @@ export function createStore(web3: Web3) {
         ]);
       },
 
-      async mintWeaponN({ state, dispatch }, {num}) {
+      async mintWeaponN({ state, dispatch }, {num, choosenElement}) {
```
seiyria, line 956: "chosen"

frontend/src/utils/common.ts
```diff
@@ -4,8 +4,6 @@ import BigNumber from 'bignumber.js';
 BigNumber.config({ ROUNDING_MODE: BigNumber.ROUND_DOWN });
 BigNumber.config({ EXPONENTIAL_AT: 100 });
 
-export const apiUrl = (url: string) => `${process.env.VUE_APP_API_URL || 'https://api.cryptoblades.io'}/${url}`;
```
seiyria, line 7: dont revert these changes

frontend/src/views/Blacksmith.vue
```diff
@@ -79,17 +79,51 @@
                 1+ star @ 100% chance.
               </div>
             </b-modal>
+
+            <b-modal hide-footer ref="forge-element-selector-modal" title="Select Element" @hide="disableConfirmButton = true, forgeCostMultiplier = null">
+              <div class="row justify-content-md-center">
+                <img v-on:click="setChoosenElement($event, 4)" class="elements-modal" src="../assets/Question_mark_white_icon.png"/>
```
seiyria, line 85: `:click` is sufficient here

frontend/src/views/Blacksmith.vue
```diff
@@ -79,17 +79,51 @@
                 1+ star @ 100% chance.
               </div>
             </b-modal>
+
+            <b-modal hide-footer ref="forge-element-selector-modal" title="Select Element" @hide="disableConfirmButton = true, forgeCostMultiplier = null">
```
seiyria, line 83: do not use the comma operator. make this a function.

frontend/src/views/Blacksmith.vue
```diff
@@ -79,17 +79,51 @@
...
+              <div class="row justify-content-md-center">
+                <img v-on:click="setChoosenElement($event, 4)" class="elements-modal" src="../assets/Question_mark_white_icon.png"/>
+                <img v-on:click="setChoosenElement($event, 0)" class="elements-modal" src="../assets/elements/fire.png"/>
+                <img v-on:click="setChoosenElement($event, 1)" class="elements-modal" src="../assets/elements/earth.png"/>
+                <img v-on:click="setChoosenElement($event, 2)" class="elements-modal" src="../assets/elements/lightning.png"/>
+                <img v-on:click="setChoosenElement($event, 3)" class="elements-modal" src="../assets/elements/water.png"/>
+              </div>
+              <div class="row justify-content-md-center" style="margin-top: 2%;">
+                <b-button
+                v-if="chlickedForgeButton === 0"
```
seiyria, line 93: "clicked"

frontend/src/views/Blacksmith.vue
```diff
@@ -79,17 +79,51 @@
...
+
+            <b-modal hide-footer ref="forge-element-selector-modal" title="Select Element" @hide="disableConfirmButton = true, forgeCostMultiplier = null">
+              <div class="row justify-content-md-center">
+                <img v-on:click="setChoosenElement($event, 4)" class="elements-modal" src="../assets/Question_mark_white_icon.png"/>
+                <img v-on:click="setChoosenElement($event, 0)" class="elements-modal" src="../assets/elements/fire.png"/>
+                <img v-on:click="setChoosenElement($event, 1)" class="elements-modal" src="../assets/elements/earth.png"/>
+                <img v-on:click="setChoosenElement($event, 2)" class="elements-modal" src="../assets/elements/lightning.png"/>
+                <img v-on:click="setChoosenElement($event, 3)" class="elements-modal" src="../assets/elements/water.png"/>
+              </div>
+              <div class="row justify-content-md-center" style="margin-top: 2%;">
```
seiyria, line 91: do not use style attributes. use CSS.

frontend/src/views/Blacksmith.vue
```diff
@@ -79,17 +79,51 @@
...
+            </b-modal>
           </div>
         </div>
 
         <weapon-grid v-model="reforgeWeaponId" />
         <b-modal size="xl" class="centered-modal " ref="new-weapons" ok-only>
                     <template #modal-header>
                          <div v-if="!spin" class="new-weapon-header-text text-center">
-                              <strong>Forging Results</strong>
+                              <strong>A-hooooy! These things look shurpppp!</strong>
```
seiyria, line 123: do not revert these text changes

frontend/src/views/Blacksmith.vue
```diff
@@ -247,7 +289,7 @@ export default {
       }, 10000);
 
       try {
-        await this.mintWeaponN({num: this.forgeMultiplier});
+        await this.mintWeaponN({num: this.forgeMultiplier, choosenElement: this.selectedElement});
```
seiyria, line 292: "chosenElement"

contracts/cryptoblades.sol
0xroy: How do I avoid this I thought I merged correctly from the main project

frontend/src/views/Blacksmith.vue
```diff
@@ -261,6 +303,25 @@ export default {
...
+    },
+
+    setChoosenElement(ele, i) {
+      this.selectedElement = i;
+      if (i === 4) {
+        this.forgeCostMultiplier = 1;
+      } else{
+        this.forgeCostMultiplier = 2;
+      }
+      [].forEach.call(ele.srcElement.parentNode.childNodes, function(child) {
```
seiyria, line 318: don't do this. cast it to an array so you don't have to call it like this. also, use arrow functions.

frontend/src/views/Leaderboard.vue
```diff
@@ -48,7 +47,7 @@ export default {
 
   async created() {
     this.waitingLeaderboardOutcome = true;
-    const leaderboardData = await fetch(apiUrl('static/leaderboard'));
+    const leaderboardData = await fetch('https://api.cryptoblades.io/static/leaderboard');
```
seiyria, line 50: revert this

seiyria (changes requested):
this does not look like it was merged successfully; a lot of changes were reverted and they must be fixed. additionally, the weapon icons look very different in size and should be consistent

contracts/cryptoblades.sol
seiyria: It looks like you did a lot of `--strategy=mine` merging, which as you can see completely reverted the last day or so of changes. I have no idea what setting you need to set to fix this though, but I would recommend re-merging your changes if possible so these overwrites don't happen.

0xroy:
The x10 Forge if you chose an element doesn't really work rn, I saw on Discord that there is a problem with the x10 so it might me related to it but the x10 with random selected works just fine.

frontend/src/views/Blacksmith.vue
```diff
@@ -80,35 +80,35 @@
               </div>
             </b-modal>
 
-            <b-modal hide-footer ref="forge-element-selector-modal" title="Select Element" @hide="disableConfirmButton = true, forgeCostMultiplier = null">
+            <b-modal hide-footer ref="forge-element-selector-modal" title="Select Element" @hide="onHideModal">
               <div class="row justify-content-md-center">
-                <img v-on:click="setChoosenElement($event, 4)" class="elements-modal" src="../assets/Question_mark_white_icon.png"/>
+                <img v-on:click="setChoosenElement($event, 100)" class="elements-modal" src="../assets/Question_mark_white_icon.png"/>
```
seiyria: this is still "choosen"

frontend/src/views/Blacksmith.vue
```diff
@@ -304,18 +304,18 @@ export default {
...
-      this.chlickedForgeButton = i;
+      this.clickedForgeButton = i;
       this.$refs['forge-element-selector-modal'].show();
     },
 
     setChoosenElement(ele, i) {
       this.selectedElement = i;
-      if (i === 4) {
-        this.forgeCostMultiplier = 1;
+      if (i === 100) {
```
seiyria, line 342: this could be `this.chosenElementFee = i === 100 ? 1 : 2;`

seiyria:
One mostly smallish change, another cosmetic change. Also, can you show me a picture of the updated UI with the icons all being the same size?

0xroy:
They elements start all with the same size with a hover effect and if you click one it transfroms to a bigger size.
https://i.gyazo.com/ca9995b2190abefc9a17ddb8e22c7940.mp4

seiyria:
I see. Sorry, can you instead add an outline effect? The bigger bit is a bit hard to tell at a glance.

0xroy:
Sorry for the late response. This is how it looks now:
https://gyazo.com/13eb83af7e58df4b8b3cac534e5013e4
Unfortunately are the contracts I wrote outdated and I might need some help rewriting them.

seiyria:
This looks pretty good, but yeah the conflicts need to get fixed.

0xroy:
I reworked the contracts accordingly but someone definitely needs to look over them.
Also the x10 Forge only worked two times then it threw an error that something in the contracts is wrong, but the x1 worked just fine.

contracts/weapons.sol
```diff
@@ -208,15 +208,15 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
             stars = 0; // 1* at 44%
         }
 
-        return mintWeaponWithStars(minter, stars, seed);
+        return mintWeaponWithStars(minter, stars, seed, chosenElement);
     }
 
-    function mintWeaponWithStars(address minter, uint256 stars, uint256 seed) public restricted returns(uint256) {
+    function mintWeaponWithStars(address minter, uint256 stars, uint256 seed, uint16 chosenElement) public restricted returns(uint256) {
```
kcper, line 214: Weapon trait is uint8 originally, I think it's better to keep it matching.

contracts/weapons.sol
```diff
@@ -245,9 +245,15 @@ contract Weapons is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
         return tokenID;
     }
 
-    function getRandomProperties(uint256 stars, uint256 seed) public pure returns (uint16) {
+    function getRandomProperties(uint256 stars, uint256 seed, uint16 chosenElement) public pure returns (uint16) {
```
kcper, line 248: Weapon trait is uint8 originally, I think it's better to keep it matching.

contracts/cryptoblades.sol
```diff
@@ -711,24 +711,24 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         }
     }*/
 
-    function mintWeaponN(uint32 num)
+    function mintWeaponN(uint32 num, uint16 chosenElement, uint16 chosenElementFee)
```
kcper, line 714: Needs check on chosenElementFee value to prevent passing 0 as chosenElementFee there and minting weapons for free with `line 720:  _payContract(msg.sender, mintWeaponFee * num * 0);`

Use e.g. `require(chosenElementFee > <whatever the required fee is>)`

contracts/cryptoblades.sol
```diff
@@ -711,24 +711,24 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
 
         for (uint i = 0; i < num; i++) {
-            weapons.mint(msg.sender, uint256(keccak256(abi.encodePacked(blockhash(block.number - 1), msg.sender, i))));
+            weapons.mint(msg.sender, uint256(keccak256(abi.encodePacked(blockhash(block.number - 1), msg.sender, i))), chosenElement);
         }
     }
 
-    function mintWeapon() public onlyNonContract oncePerBlock(msg.sender) {
-        _payContract(msg.sender, mintWeaponFee);
+    function mintWeapon(uint16 chosenElement, uint16 chosenElementFee) public onlyNonContract oncePerBlock(msg.sender) {
```
kcper, line 727: Same chosenElementFee vulnerability here.

0xroy: Issues are resolved accordingly

contracts/cryptoblades.sol
```diff
@@ -711,20 +711,25 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         }
     }*/
 
-    function mintWeaponN(uint32 num, uint16 chosenElement, uint16 chosenElementFee)
+    function mintWeaponN(uint32 num, uint8 chosenElement)
```
kcper, line 714: I just noticed and I think we should only allow passing 0, 1, 2, 3, 100 as chosenElement, other values are not supported and could cause unexpected behavior.
Please create a modifier to check for supported elements and apply to both mint functions.
Something along those lines:
```
modifier isSupportedElement(uint8 element) {
    _isSupportedElement(element);
    _;
}

function _isSupportedElement(uint8 element) internal view {
   require(element == 100 || (element>= 0 && element<= 3), "Not supported element");
}
```

I'm not too sure what effect it will have on the gas cost - if it's significant it might be better to create a separate function for mints with chosenElement and leave this as it was originally for random elements. @poshdan please share your opinion on this.

poshdan: gas costs are less important for things like minting - daily we have 3 million fight transactions but weapon mints are only in the thousands (let's say 50k weapons made, and that's not accounting for how many weapons are made via 10x).

That would make once cent saved on fights 60x more important than one cent saved in mints.
Of course there's no need to be wasteful, but I'm not going to lose sleep over either approach.
The tradeoff with two functions instead of one, is that the contract becomes larger, and we are indeed close to hitting limits. We can switch it around later if we need to compress the contract though.

Feel free to try the function before and after in your local environment or run tuffle console's estimate function like so:
mintWeapon.estimateGas(parameters of mintWeapon). Multiply whatever gas cost you get by x1.45 (which is gonna be roughly its cost on mainnet), and at 5 gwei it takes 200 million gas per BNB. If the difference in gas is less than several thousand, it's not a big deal.

0xroy: With the modifier:
image

With the separate functions:
image

I personally favor the modifier solution because its much less code and thus easier to read but you guys decide.

kcper: Nice, go for the modifier definitely then. Thanks!

0xroy: Alright updated it thanks for your help guys!

kcper (approved):
Looks good, thanks. x10 forging works as expected now?

0xroy:
Yes no problem with it anymore works like a charm.

poshdan:
I tested it locally, seems good. This should be ready to merge but needs a migration file (which we can take care of, during the next deploy, so don't worry about that)

raymond-h (approved):

PR #589, opened 2021-08-04
Add shields.sol

brian-man:


poshdan:
I did some work on this to make them purchasable.
requestPayFromPlayer is notoriously missing from the function as it takes int128 USD instead of a converted value. There's an ticket somewhere about moving that require statement into the pay function itself anyway.

Plugged in the purchase thing to store.ts. To get the remaining supply we'd have to call the shields contract and do "10000-totalSupply()" to display that in the frontend, but hooking up that contract requires quite a bit of extra frontend work. (which will be needed to display the list of owned shields anyway) A temporary workaround would be to have a view function in blacksmith.sol that does the totalsupply calculation.

raymond-h (approved):

Issue #477, opened 2021-07-27
429 detection


seiyria:
on the marketplace and leaderboard page, there needs to be a notice when a 429 is returned instead of the normal data expected

PR #593, opened 2021-08-04
477-detect-too-many-requests

bilbolPrime:
### All Submissions

* [ ] 
image

### New Feature Submissions

* [ ] #477 

### Notes

### PR Description

A simple check on response code in Leaderboards and market. I think it's a good idea to open an enhancement issue to have a centralized API handler or so but I hope this is sufficient to close this issue. Went with 1 minute to avoid cooldown reset if people try to exactly match whatever threshold is active.

**Note:** Could not properly test with API locally. Used https://httpstat.us/429 to simulate. Market APIs need to be validated on testnet.


frontend/src/views/Leaderboard.vue
```diff
@@ -49,6 +49,11 @@ export default {
   async created() {
     this.waitingLeaderboardOutcome = true;
     const leaderboardData = await fetch(apiUrl('static/leaderboard'));
+
+    if(leaderboardData.status === 429){
+      this.$dialog.notify.error('You are making too many requests. Please try again in 1 minute.');
```
seiyria, line 54: there should be a return here to bail as well

seiyria (approved):

bilbolPrime:
@seiyria I create a new helper class (APIHandledCall) that handles the fetch and triggers delegated functions based on response code.
This should be the cleanest way to allow halting without throwing exceptions (success callback only called if we got a 200 response).

Ideally the 429 dialog alert would be called inside the class without the need to delegate anything but that would require passing this reference and some imports so I just left it as an optional delegate for the caller.

Is this sufficient or do you want me to add more methods to the class (example addParameter etc..)?

Will do full cycle testing before making this a live PR but unless I encounter a bug while testing or you ask me to expand the class this should be it.

frontend/src/views/Leaderboard.vue
bilbolPrime: Done in the new approach (delegate a success callback)

frontend/src/utils/api-helper.ts
```diff
@@ -0,0 +1,64 @@
+const OK_RESPONSE: number = 200;
+const TOO_MANY_REQUESTS_RESPONSE: number = 429;
+
+export class APIHandledCall {
```
seiyria, line 4: honestly this is a bit large of an abstraction. we're not going to need onFail, always, onTooManyRequests etc. What I would do instead is have an `APIHelper` that has `get`, taking a URL, and always doing the same thing for 429 etc.

this is a bit bulky. if possible, this should use an http interceptor for maximum usability: https://stackoverflow.com/a/37252819/926845

seiyria (changes requested):
see comment

frontend/src/utils/api-helper.ts
bilbolPrime: I agree that this may be too much. The reason I didn't go with an interceptor is because of the bailing part (stop execution of code because no actual result returned). I don't think the interceptor can do that, only redirect completely.

I will try to do a best of both worlds; the interceptor to show 429 warning and a APIHelper with a onsuccess callback. Or perhaps a simple inline check for 200 success response and drop the APIHelper alltogether.

I'll see what I can cook on Wednesday.

seiyria: I'm not sure we need an APIHelper for something like that. A typical HTTP call would be:

- make call
- check if data is present
- if no data, bail

That's how it should currently be anyway. A 429 interceptor is a good idea (that's the idea behind interceptors, is to process data on the outgoing or incoming side, but nothing else)

bilbolPrime:
Putting this here as this seems to be a good reference
https://www.codewhoop.com/blog/vuejs/vue-resource-interceptors-with-typescript.html

Kept running into issues; http not defined even though I did the "augmenting types" thing. The vue as any work around didn't' work either; gave interceptor undefined.

Also adding vue-resource gave an error on build.

Some Vue 2 / 3 issues or I am doing it very wrong even though it was copy paste from that site. Based on my front end skill it's probably the latter.

Will give this another shot later as we have more pressing priorities (market).

PR #596, opened 2021-08-04
Earning Calculator fix

Neuche:
###  Description

New features for the Earning Calculator

- Added stamina selector
- Added tooltip

### Notes
Earnings are calculated in a very confusing way, I took the gas offset away from power reward calculation, and reimplemented it depending on the number of fights, which depend on the stamina selector.

Calculator now considers a 'day' as 7.2 fights, which will essentially mean bigger numbers overall for the users.

### Screenshots

image

-Tooltip
image

### DIFF
--- a/frontend/src/components/smart/EarningsCalculator.vue
+++ b/frontend/src/components/smart/EarningsCalculator.vue
@@ -31,14 +31,9 @@
                   <b-form-input class="stat-input" type="number" v-model="levelSliderValue" :min="1" :max="255" />
                 </div>
                 <span>Stamina</span>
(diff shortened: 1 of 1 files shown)


seiyria:
this diff is pretty big.

concerns:
- this is not merged with the latest main. the top area should be labels, not inputs
- all inputs should have `.form-control` applied (this extends to more than yours, of course, but I'd rather not continue to perpetuate this)
- that tooltip needs to read much differently, the random caps everywhere are not ideal I'm not sure the tooltip needs to even be there as it's supposed to be the same, no matter how many times you do it.

Issue #208, opened 2021-07-10
Show actual gas spent in fight results


crypto-smoke:

cpp-vortx:
is this being worked on?

seiyria:
nope! all yours.

cpp-vortx:
Alright, later I'm gonna take it with my personal git account, this is actually my corporate account

caiquejjx:
Gonna start looking into this, can u assing me into the issue?

seiyria:
assignments can't be made until you're part of the github org

caiquejjx:
the gas spent should be shown in BNB right?

seiyria:
@caiquejjx yes

PR #601, opened 2021-08-04
feat(combat): show gas result post-combat

caiquejjx:
### All Submissions

* [X] Can you post a screenshot of your changes (if applicable)?

WIN:
image

LOSE:
image

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/208

### Notes

### PR Description
Added gas used info in combat results

### DIFF
--- a/abi-common.ts
+++ b/abi-common.ts
@@ -12,6 +12,7 @@ export interface Web3JsTransactionResult {
   events: any;
   blockNumber: number;
   gasUsed: number;
+  transactionHash: string;
 }
 
 type BlockNumber = number | string | 'latest' | 'earliest' | 'pending';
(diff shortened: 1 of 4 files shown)


caiquejjx:
@seiyria based on that site https://bscgas.info I used the gas price as 5 since most of the time the price is 5 and I believe this feature is supposed to be an approximation but I could easily change to an api call that gets the current BSC gas price prediction, do u think it's necessary to use the api?

ed3ath:
use web3.eth.getTransaction() to get the gas price and web3.eth.getTransactionReceipt() to get the gas used

that's what I used on my tracker. it's 100% accurate

caiquejjx:
The getTransaction might be a good idea for the gas price but the getTransactionReceipt I don't think it's necessary since the transaction itself already returns the gas used

ed3ath:
> The getTransaction might be a good idea for the gas price but the getTransactionReceipt I don't think it's necessary since the transaction itself already returns the gas used

yeah. so all good? :D

caiquejjx:
yep, gonna implement the getTransaction and commit, thanks!

seiyria (approved):

seiyria (approved):

PR #612, opened 2021-08-05
trait icon conflict css w/ shield icon adjustment

wintersummermint:
### All Submissions

alignment trait icon issue

### New Feature Submissions

### Notes

### PR Description

(insert description here)


byyyyyyyyy:
can i get a pr?

byyyyyyyyy (commented):
i have a suggestion for this code change. I think the culprit is inside the NftIcon.vue component. just add scoped attribute in the style. no need to add another class on this file. that would fix it i think. already working on my local. hope this helps. can correct me also if i'm not doing it right.

seiyria:
good call @byyyyyyyyy - that worked and is imo a better solution as we should have scoped styles for everything (except what we intend to be global)

PR #616, opened 2021-08-05
ERC20 recover ability to blacksmith

poshdan:
A couple of users mistakenly sent 5 skill to the blacksmith contract in hopes that it would mint a shield for them.
This function will probably come in handy in the future too, seeing as the blacksmith will be used for other paid actions too.
I will kindly accept the 10 skill I paid out of pocket as my bounty


poshdan:
I got refunded for this recently so I'm not terribly invested in it now - though we do have 10 skill stranded on the contract and may have more soon, so it's still probably worth adding this. There's also like a hundred BUSD there for some reason.

raymond-h (approved):

PR #621, opened 2021-08-06
Remove xp updating from mintWeapon and mintWeaponN in store.ts

brian-man:


brian-man:
@seiyria It seemed like this was/is unnecessary updating logic after weapon minting.  If not the case, I'm happy to close/abort this PR.

seiyria:
@brian-man I'm not sure sunset u had no idea this pr existed. If it's necessary I can merge it and if not close it. I need insight on what to do with it though

PR #624, opened 2021-08-06
Change seed handling for mintWeapon and mintWeaponN

brian-man:


contracts/cryptoblades.sol
```diff
@@ -451,6 +460,60 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
+        mintPayments[msg.sender].count = count;
+        mintPayments[msg.sender].nftAddress = nftAddress;
+        mintPayments[msg.sender].blockNumber = block.number;
+        mintPayments[msg.sender].blockHash = 0;
+    }
+
+    function _usePayment(address nftAddress, uint count) internal {
+        _discardPaymentIfExpired();
+
+        require(mintPayments[msg.sender].count >= count);
```
poshdan, line 506: we may want the .count check to be == instead of >=.
Imagine paying for 10 mints, then for some reason deciding to use the 2nd step for the 1x mint.

contracts/cryptoblades.sol
```diff
@@ -451,6 +460,60 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
+        _payContract(msg.sender, mintWeaponFee);
+
+        mintPayments[msg.sender].count = count;
+        mintPayments[msg.sender].nftAddress = nftAddress;
+        mintPayments[msg.sender].blockNumber = block.number;
+        mintPayments[msg.sender].blockHash = 0;
+    }
+
+    function _usePayment(address nftAddress, uint count) internal {
+        _discardPaymentIfExpired();
```
poshdan, line 504: I wonder if the discardIfExpired call is necessary if the try portion is to succeed

contracts/cryptoblades.sol
```diff
@@ -451,6 +460,60 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
+            // Payment must be recent enough that the hash is available for the payment block.
+            // Use 200 as a 'friendly' window of "You have 10 minutes."
+            if (mintPayments[msg.sender].blockNumber + 200 >= block.number) {
+                mintPayments[msg.sender].blockHash = blockhash(mintPayments[msg.sender].blockNumber);
+            }
+        }
+    }
+
+    function _discardPaymentIfExpired() internal {
+        _updatePaymentBlockHash();
```
poshdan, line 477: we probably want to do the blockhash update in the catch part of the minting code?

contracts/cryptoblades.sol
brian-man: > we may want the .count check to be == instead of >=.
> Imagine paying for 10 mints, then for some reason deciding to use the 2nd step for the 1x mint.

With >= the code already supports this scenario.  Changing it to == breaks that.

contracts/cryptoblades.sol
brian-man: > I wonder if the discardIfExpired call is necessary if the try portion is to succeed

I'm not understanding.  The general idea is that discard actually does two things: (1) discard expire payments and (2) if the payment is present, not expired, and we can update block hash do so.

We could split the discard and make two calls, but that seemed unnecessary.

I'm not clear on your thought here though.  Elaborate?

contracts/cryptoblades.sol
brian-man: > we probably want to do the blockhash update in the catch part of the minting code?

I was thinking we want to update the blockhash as soon as possible.

```
    function mintWeapon() public onlyNonContract oncePerBlock(msg.sender)  {
        _discardPaymentIfExpired();
        try weapons.mint(msg.sender, randoms.getRandomSeed(msg.sender)) {
            _usePayment(address(weapons), 1);
        }
        catch {
        }
    }
```

My understanding is that the effects of _discardPaymentIfExpired() will not be reverted if the try fails.

If that's not the case, then we need to add redundant-looking logic to the catch:

```
    function mintWeapon() public onlyNonContract oncePerBlock(msg.sender)  {
        _discardPaymentIfExpired();
        try weapons.mint(msg.sender, randoms.getRandomSeed(msg.sender)) {
            _usePayment(address(weapons), 1);
        }
        catch {
            _discardPaymentIfExpired();
        }
    }
```

raymond-h (approved):

PR #644, opened 2021-08-08
Weapon hashing no longer has 10 minute drama

brian-man:


contracts/cryptoblades.sol
```diff
@@ -144,28 +139,13 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
-
-        uint256 skillRefundableFromWallet;
-        uint256 skillRefundableFromRewards;
-        uint256 skillRefundableFromIgo;
-
-        uint256 refundClaimableTimestamp;
-    }
-
-    uint256 public totalMintPaymentSkillRefundable;
-    mapping(address => MintPaymentSkillDeposited) mintPaymentSkillDepositeds;
```
raymond-h, line 168: Removing these things will make the contract not backwards-compatible, so they must be left in.

raymond-h (changes requested):

raymond-h:
Fails to compile. Seems to be because IRandoms and ChainlinkRandoms were modified, but DummyRandoms was not modified to match.

brian-man:
@raymond-h The dummy randoms is probably a trivial thing to correct, but the bigger issue as that we can't seem to stop mucking with minting in other ways and it was 8 days between updated PR and review attention.  It's just not sustainable for me to keep merging against a moving target with a week+ turnaround for review time.

Issue #119, opened 2021-06-26
skill shop


seiyria:
There should be a new shop tab that lets users buy items with skill. It should have some of these (balance pending):

- [x] a potion of rename (to rename a single character) for 10 skill (unlimited buys)
- [ ] a potion of experience gain (+10% XP for Y days for all characters) for Z skill (X / week)
- [ ] a potion of power that reduces the spread of enemy power from +/- 10 to +/- 5 for X skill (Y / week)
- [ ] a potion of weakness that reduces enemy power to 1 for a single fight for X skill (Y / week)
- [ ] ~~a potion of stamina to refill stamina for one character for X skill (Y / week)~~
- [ ] a potion of boosted strength (+10% power to all combats for X days) (Y / week)

All shop stuff should reset on probably friday nights.

We also want to be able to add stuff to shops on a biweekly basis, leaving the shop in two fridays. Some examples would be:

- [ ] rotating titles which apply to your wallet
- [ ] a limited time weapon, possibly a weak 5* or strong 4*, for X skill. ideally with a prefab construction so we can control the aesthetics
- [ ] limited time character cosmetics for X skill
- [ ] get a cosmetic for your character, such as a unique border (int on model, all done with css)

Giovanny-DS:
a potion that changes the element of a character.

this would be useful for new. if were encouraging to grow our chars instead of creating an army of LVL 1 - 21 accounts. 

Usecases:

well to new starters, they can go all the way without having to buy in the market a higher prices chars of the same element. instead, they can recruit the four chars (which is cheaper) and **EARN** then they can turn all of the same elements at the same time they earn.

also if a player does need, (whatever the reason is) a high-level char of an element, they can change the element of that char without worrying about trading or selling it to buy another.

this will also provide a better game experience for those players who entered without that much knowledge and recruited their four chars. they can fix or adapt their team to any situation or weapon they have.

as an item that can be bought with skill. this will also help the economy getting the back skills to the contract. even if it's not used that much.

brian-man:
> a potion of weakness that reduces enemy power to 1 for a single fight for X skill (Y / week)
Kind of weird to have an "I win" potion.

Will the payout be proportional to 1 monster power? :)

brian-man:
> a potion that changes the element of a character.

Agree, a trait changer is simple to implement and doesn't unbalance the game.  With the right price it could be a huge skill sink prior to raids.

seiyria:
[...]

PR #659, opened 2021-08-12
fix(ui): run censor on names to filter profanity

kcper:
### All Submissions

image
image
image
image

### New Feature Submissions

Follwup to #119 
### Notes

### PR Description

Censoring everywhere. Check package .jsons, because my changes wanted to delete 27k lines and I cleaned it manually to leave only relevant stuff.


frontend/src/components/WeaponIcon.vue
```diff
@@ -88,6 +88,8 @@ const gColor = new Three.Color(0x413F41);
 const bColor = new Three.Color(0xAF5822);
 const white = new Three.Color(0xFFFFFF);
 
+const censor = new CensorSensor();
```
seiyria, line 91: can this get abstracted to a separate service or somewhere else it can be imported? like the getWeaponNameFromSeed function etc. Then we have one function we can call everywhere instead of initializing this everywhere. If we want to update the profanity we then also have one file we have to do so.

seiyria (changes requested):
abstract censorSensor

kcper:
Added abstract instance. Fixed renames not being loaded in the market, I could see it earlier, because it was my nft and rename was loaded after renaming.

PR #672, opened 2021-08-14
load market taxes when switching to search tab

### LINKED ISSUE #665: NaN in price for sell dialog
image

kevinunger:
### All Submissions
Fixes this bug. (fixes issue 665)
image

### PR Description

This fixes #665. The market tax is not set when not coming from the listing tab, because that's the only time this.weaponMarketTax is set. This fix loads the market tax, when selecting the 'Search NFTs' tab, by calling loadMarketTax().

### DIFF
--- a/frontend/src/views/Market.vue
+++ b/frontend/src/views/Market.vue
@@ -195,7 +195,7 @@
         </div>
       </b-tab>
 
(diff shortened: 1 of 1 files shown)


seiyria (approved):
Multiple calls really should not be made in the same dom line, but you didn't do it so it's fine. But keep it in mind.

Issue #674, opened 2021-08-15
Cookie consent banner for all sites


kevinunger:
We need a consent banner when using ads to comply with data-protection regulations like GDPR.

seiyria:
We don't collect any data besides what Google analytics collects.

kevinunger:
It's still needed under EU-Law, because services like analytics and adwords are collecting user data and using cookies from the users browser.

kevinunger:
Don't have time to do that today. If someone else wants, go for it, please. The loading of google analytics in index.html must be disabled, before the user consents.

PR #686, opened 2021-08-20
feat(analytics): add cookie consent banner

Sellie96:
image

On Reject Cookies aren't saved for the user on visits

image

On Accept Cookies persist between sessions

image

#674 

Added a Consent banner for the use of google analytics to display targeted ads and measure site data.


frontend/public/index.html
```diff
@@ -44,11 +37,37 @@
...
+    <p>
+    </p>
+  </div>
   <noscript>
     <strong>We're sorry but CryptoBlades doesn't work properly without JavaScript enabled.
       Please enable it to continue.</strong>
   </noscript>
   <div id="app"></div>
+  <link rel="stylesheet" href="//rawcdn.githack.com/Alex-D/Cookies-EU-banner/2.0.0/css/cookies-eu-banner.default.css">
+  <script src="//rawcdn.githack.com/Alex-D/Cookies-EU-banner/2.0.0/dist/cookies-eu-banner.min.js"></script>
```
seiyria, line 58: is there a way we can install this instead of referencing someone elses code?

frontend/public/index.html
```diff
@@ -44,11 +37,37 @@
...
+  <script src="../node_modules/cookies-eu-banner/dist/cookies-eu-banner.min.js"></script>
```
seiyria, line 59: can this be referenced by `import 'cookies-eu-banner'` in a main file, instead of importing from node modules?

seiyria (changes requested):

JustinSibonga (approved):

seiyria:
@JustinSibonga who are you?

kevinunger:
I copied the scripts/css from node_modules to public to load it from there. 
I also moved the loading of the google-ad script into the cookiesEuBanner function, to load it only when consent has been given.
When the google-ad script is loaded outside of that (like it was before) it's always setting the _gads cookie. 

In the future, we might use a vue plugin for analytics and ads. This way we can just do it the vue-way and use analytics and ads in the components more easily. This also enables event-tracking for GA, so more granular tracking can be done in the future.

kevinunger:
resolves #674

seiyria (approved):

PR #690, opened 2021-08-21
Fix canJoinRaid() in raid1.sol

brian-man:


poshdan:
Superseded by Raid UI patch #694 as it seems to extend on it.
Bounty should still be upheld for this one (it was developed in parallel afaik)

brian-man:
> Superseded by Raid UI patch #694 as it seems to extend on it.
> Bounty should still be upheld for this one (it was developed in parallel afaik)

Sort of.  Taloner wasn't aware this was already in review so he duplicated the effort after the fact.

PR #691, opened 2021-08-21
Stop deduction and transfer of zero in _deductPlayerSkillStandard()

brian-man:
Noticed when looking at `mintWeapon()` scraping results that showed transfers of zero (ERC-20) Skill.


brian-man:
@ed3ath I tried to add you as a second assignee but the UI (top right) is being finicky.

raymond-h (approved):

Issue #490, opened 2021-07-28
transak integration


seiyria:
https://www.notion.so/Integrate-Transak-f43cbb5dc7fa4d068a044ef9589879bf

will be used to buy bnb

EduardoGomez951020:
Sharip I'll take it

EduardoGomez951020:
Please add me to org.


![transak integration ](https://user-images.githubusercontent.com/28773860/130180505-7cd89ebd-5c9b-4cbf-a301-af88523b7257.jpeg)

seiyria:
@EduardoGomez951020 I add people after their first contributionb is merged and I think this is your first still

PR #695, opened 2021-08-21
transak integration #490

EduardoGomez951020:
Adding Modal for buy SKILL and transak

Screen Shot 2021-08-21 at 12 20 06


frontend/src/store.ts
```diff
@@ -241,7 +241,9 @@ export function createStore(web3: Web3) {
       getExchangeUrl() {
         return 'https://app.apeswap.finance/swap?outputCurrency=0x154a9f9cbd3449ad22fdae23044319d6ef2a1fab';
       },
-
+      getExchangeTransakUrl() {
+        return 'https://global.transak.com/?apiKey=YOUR_PRODUCTION_API_KEY?defaultCryptoCurrency=BNB'; //INSERT API KEY TRANSAK
```
seiyria, line 245: this should be marked as draft if we don't have an api key

seiyria (changes requested):
needs an api key

EduardoGomez951020:
Hi @seiyria. Where or how should I get the api key ? 

Is existing already on the .env file ? 

Screen Shot 2021-08-23 at 11 01 54

seiyria:
@EduardoGomez951020 I have requested one from kroge, but as of right now there isn't one.

EduardoGomez951020:
Hi @seiyria, any update about it ?

seiyria:
looks fine, it took a while to get the api key (it looks like they forgot about us). sorry for the delay. this needs a round of testing then it would be good to go if it all works.

bilbolPrime:
Created PR 742 with some modifications to this one. Had trouble cloning the repo.

PR #697, opened 2021-08-22
Resolves #696

### LINKED ISSUE #696: Confirm reforge button with dust should be disabled, when no dust is selected
This should be disabled:
image

kevinunger:
Button is now disabled, when no dust is selected for reforge:
image

Enabled, when any dust is selected:
image


kcper (approved):
Thanks :)

PR #708, opened 2021-08-28
Avoid transferring zero in cryptoblades.sol

brian-man:


raymond-h (approved):

kcper:
Outdated, only `recoverSkill` change is valid here.
Currently all other transfers that were modified here are handled with `_deductPlayerSkillStandard` which checks `if(fromUserWallet > 0) {`

PR #711, opened 2021-08-28
Raid.vue rewrite to Typescript

Chriserus:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

My PR includes changes made to Raid.vue file (and some additional files), my changes are there in order to rewrite and take advantage of Typescript features. In addition to typical rewrite to Typescript I also found and fixed a potential bug caused by using a wrong field on RaidRewards object (keybox instead of keyboxes). My introduction of RaidRewards Interface ensures correct field usage.

### DIFF
--- a/frontend/src/views/Raid.vue
+++ b/frontend/src/views/Raid.vue
@@ -190,7 +190,7 @@ import BigButton from '../components/BigButton.vue';
 import WeaponIcon from '../components/WeaponIcon.vue';
 import Hint from '../components/Hint.vue';
 import NftIcon from '@/components/NftIcon.vue';
-import NftList from '@/components/smart/NftList.vue';
+import NftList, { NftIdType } from '@/components/smart/NftList.vue';
 import CurrencyConverter from '../components/CurrencyConverter.vue';
 import { GetTotalMultiplierForTrait, IWeapon } from '@/interfaces/Weapon';
 import { CharacterPower, IRaidState, IState } from '@/interfaces';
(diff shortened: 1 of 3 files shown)


seiyria (changes requested):
my gut instinct is that there's too much `:any` use here. reducing that as much as possible would be ideal.

Chriserus:
> my gut instinct is that there's too much `:any` use here. reducing that as much as possible would be ideal.

Thanks for the remark! I updated the branch with new changes, the only "any" left is the one related with "this" keyword and our alert system (it's done as in other places in the project).

seiyria (approved):

frontend/src/views/Raid.vue
```diff
@@ -385,89 +423,89 @@ export default {
...
-    async claimRewardIndex(rewardIndex) {
+    async claimRewardIndex(rewardIndex: string): Promise<void> {
       this.bonuxXpCharacterNames = [];
       this.bonuxXpAmounts = [];
       const result = await this.claimRaidRewards({
         rewardIndex
       });
 
-      const nfts = [];
+      const nfts: Nft[] = [];
```
kcper, line 446: This type should be NftIdType[]. Actual Nfts are fetched in NftList component based on id and type. While this will work as is, I think it might introduce some confusion.

Here's a prop in NftList.vue that the collection created here is passed to:
```
nftIdTypes: {
  type: Array as PropType<NftIdType[]>,
  default() {
    return [];
  },
},
```

kcper (approved):

Issue #705, opened 2021-08-26
cosmetic borders and glows for weapons


seiyria:
- there will be a mapping to track what X has what cosmetic
- there will be borders in the cosmetic shop
- there will be new dropdown entries to apply cosmetics if the items are held

some cosmetic examples: https://plnkr.co/plunk/XyASQBmlHkVmhJFukWtQ

some will be borders, others will be underlays

bilbolPrime:
For solidity implementation, I created a new contract "Cosmetics". These may be "consumables" but we need to spin off the existing consumable contract as it's an overkill to have one contract per cosmetic since it's really just an enumeration and no per cosmetic implementation like per consumable .

Cosmetics.sol: https://github.com/CryptoBlades/cryptoblades/blob/705-cosmetics-shop/contracts/Cosmetics.sol

How many cosmetics each wallet has. uint32 should be more than enough
` mapping(address => mapping(uint32 => uint32)) public owned;`

The cosmetic ids or enumeration that actually exist
`mapping(uint32 => bool) internal _cosmeticAvailable;`

A "helper" var to know when to stop iterating when getting all cosmetics owned
`uint32 internal _maxCosmeticId;`

The tracked events are

    event CosmeticGiven(address indexed owner, uint32 cosmetic, uint32 amount);

    event CosmeticUsed(address indexed owner, uint32 cosmetic, uint32 amount);

    event CosmeticRestored(address indexed owner, uint32 cosmetic, uint32 amount);

    event CosmeticGivenByAdmin(address indexed owner, uint32 cosmetic, uint32 amount);

    event CosmeticTakenByAdmin(address indexed owner, uint32 cosmetic, uint32 amount);`

WeaponCosmetic.sol: https://github.com/CryptoBlades/cryptoblades/blob/705-cosmetics-shop/contracts/WeaponCosmetic.sol

Mappings:
`mapping(uint256 => uint32) public appliedCosmetics;`

The tracked events are:

     event WeaponCosmeticApplied(address indexed owner, uint256 indexed weapon, uint32 cosmetic);`

     event WeaponCosmeticRemoved(address indexed owner, uint256 indexed weapon, uint32 cosmetic);


Will work on the blacksmith purchase and "hello world" UI implementation next

bilbolPrime:
Some changes made to solidity (dropped the max id to avoid unneeded complexity)

From solidity part it's pretty much finished. Some minor changes will be needed based on initial cosmetics deployment.

[...]

PR #720, opened 2021-09-01
705 cosmetics shop

kcper:
### All Submissions

image
image

### New Feature Submissions

#705 
### Notes

### PR Description

Introduces weapon and character cosmetics in the Skill Shop. Cosmetics can be bough with real skill only (unclaimed, wallet). Cosmetics can be applied one at a time via dropdown in upper right corner of each NFT. Cosmetics are reusable, applying a new one takes off the current one.


seiyria (approved):

poshdan:
I would rather we keep using the existing mappings for items.
itemAddresses instead of cosmeticAddresses
etc

bilbolPrime:
> 
> 
> I would rather we keep using the existing mappings for items.
> itemAddresses instead of cosmeticAddresses
> etc

Gave the cosmetics id 7 and 8.

@kcper please share any concerns you may have with the changes added. Did not create a new migration script so a fresh chain is needed in case you want to run it locally.

@seiyria very minor change done to UI (contract.ts; how we get the address of contract) so won't re-request review from you but please check in case you may have any concern.

frontend/src/components/smart/CharacterList.vue
```diff
@@ -142,13 +163,29 @@ export default {
...
       targetTrait: '',
       currentCharacterId: null,
-      options: []
+      options: [],
+      haveCharacterCosmetic1: 0,
+      haveCharacterCosmetic2: 0,
+      haveCharacterCosmetics: [0],
+      targetSkin: '',
+      showCosmetics: true,
+      characterCosmeticsNames: [
```
Chriserus, line 172: I would suggest to extract it to enum and use it here, this way it can be modified in one place if it would be used someplace else.

frontend/src/components/smart/WeaponGrid.vue
```diff
@@ -223,7 +246,20 @@ export default Vue.extend({
       options: [],
       haveRename: 0,
       weaponRename: '',
+      haveWeaponCosmetics: [],
+      targetSkin: '',
       currentWeaponId: null,
+      weaponCosmeticsNames: [
```
Chriserus, line 252: Same as with characterCosmeticsNames array

Chriserus (approved):
Some minor remarks, you can do them, or not, depending on time, nothing is out of the ordinary, looks good :)

poshdan (approved):

Issue #739, opened 2021-09-15
Land sale


kcper:
- [ ] Add land sale items in skill shop's 'Specials' section: T1, T2, T3 land
- [ ] Tier 1 land - buy a land in random location
- [ ] Tier 2 land - popup a map made of 50x50 chunks -> select a chunk -> popup confirmation containing chunk coordinates, element, land tier and buy button -> buy land in a random location of that chunk
- [ ] Tier 3 land - same as tier 2
- [ ] Chunks highlighted on hover
- [ ] Each user can buy only 1 land
- [ ] Only one tier 3 available per chunk
- [ ] Only 1mil total lands can be bought and all tier 3 lands need to be part of that list
- [ ] extra: checkbox in map modal to show chunk population, if enabled highlight chunks in green/yellow/red based on amount of lands sold in the chunk

Up to date map can be found in `#lore` discord channel.

bilbolPrime:
Added solidity support for feature:

ChunkId = x coordinate + 50 * y coordinate; coordinates started at (0, 0)

**BlackSmith:**
- cbkLandSale: reference to the CBKLandSale contract
- purchaseT1CBKLand(uint256 paying): purchase a tier 1 land
- purchaseT2CBKLand(uint256 paying, uint16 chunkId): purchase a tier 2 land
- purchaseT3CBKLand(uint256 paying, uint16 chunkId): purchase a tier 3 land
- getCBKLandPrice(uint8 tier): gets price of land in wei

**CBKLandSale:**

- salesAllowed(): check if land can be purchased.
- getAllowedLandOffset(): variable used to check if a chunk can sell T2 land.
- checkIfChunkAvailable(uint8 tier, uint16 chunkId): check if chunk is available for tier. Should be only called when user selects a chunk for T2 or T3 purchase or before a T1 purchase is requested. If T1 purchase, chunkId is a dummy variable.
- checkChunkReserved(uint16 chunkId): check if chunk is reserved
- checkChunkEstimatedPopulation(uint16 chunkId): _projected_ chunk population. Should NOT be called for every chunk. APIs to be added.
- getAvailableLand(): returns (available T1, available T2, available T3)
- getAvailableLandPerChunk(): returns available land per chunk
- getSoldLand(): returns (sold T1, sold T2, sold T3)
- getPurchase(): returns purchase if any, (purchased tier, chunkId or stamp if T1)


Will work on APIs next.

Code on 739-land-sale and draft PR opened #746

[...]

PR #746, opened 2021-09-19
WIP: 739 land sale

bilbolPrime:
### All Submissions

### New Feature Submissions

* [x] #739 

### Notes

### PR Description

To be added


contracts/CBKLandSale.sol
```diff
@@ -0,0 +1,391 @@
...
+
+    function getSalesCount() public view returns (uint256){
+        return totalSales;
+    }
+
+    function getPurchaseBySale(uint256 sale)  public view returns (address, uint8, uint32) {
+        return (sales[sale].buyer, sales[sale].purchasedTier, sales[sale].stamp);
+    }
+
+    function toggleSaleAllowed(bool allowed) external isAdmin {
```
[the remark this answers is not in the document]
bilbolPrime: this has been done in the latest push

migrations/91_cbkLandSale.js
```diff
@@ -0,0 +1,23 @@
...
+const ExperimentToken = artifacts.require("ExperimentToken");
+const CBKLand = artifacts.require("CBKLand");
+module.exports = async function (deployer, network, accounts) {
+
+  const cbkLand = await deployProxy(CBKLand, [], { deployer });
+  const cbkLandSale = await deployProxy(CBKLandSale, [cbkLand.address], { deployer });
+ 
+ 
+  const blacksmith = await upgradeProxy(Blacksmith.address, Blacksmith, { deployer });
+  const experimentToken = await ExperimentToken.deployed();
```
poshdan, line 14: correct me if I'm wrong but AFAIK we never use experiment token on any mainnets, we'd have to hardcode the address of king for each network in this migration to pass its reference instead.
Some other migrations use these with an if statement checking for dev network, so we can keep it operational on local

contracts/CBKLand.sol
```diff
@@ -0,0 +1,75 @@
...
+        require(hasRole(GAME_ADMIN, msg.sender), "Not game admin");
+    }
+
+    function get(uint256 id) public view returns (uint256, uint256, uint256, uint256) {
+        return (landData[id][LT], landData[id][LC], landData[id][LX], landData[id][LY]);
+    }
+
+    // tier, chunkid, x, y
+    function getOwned(address owner) public view returns (uint256, uint256, uint256, uint256) {
+        uint256 id = tokenOfOwnerByIndex(owner, 0);
```
poshdan, line 50: i assume this is work in progress but in case it's not  I'd point out this kinda ssumes you will only ever own one land (maybe that's the intention? let me know)

contracts/CBKLand.sol
```diff
@@ -0,0 +1,75 @@
...
+        return get(id);
+    }
+
+    // DO NOT call directly outside the logic of CBKLandSale to avoid breaking tier and chunk logic
+    function mint(address minter, uint256 tier, uint256 chunkId) public restricted {
```
poshdan, line 55: double checking here, some calls are placed with chunkId of 0, do valid chunk Ids start at 1, is 0 a placeholder of "to be determined", perhaps some logic needs to be put here to actually determine where? Come to think of it, with the scale of the map, it may be too gassy to resolve in the mint function itself, if we do it manually after via admin functions we skip over random exploitation or checking for collision on existing chunks

contracts/CBKLandSale.sol
```diff
@@ -0,0 +1,386 @@
...
+
+        if(tier == TIER_TWO){
+            return availableLand[TIER_TWO] > 0
+                    && chunkT2LandSales[TIER_TWO] < _allowedLandSalePerChunk
+                    && _chunkAvailableForT2(chunkId);
+        }
+
+        if(tier == TIER_THREE){
+            return availableLand[TIER_THREE] > 0
+                    && chunkT2LandSales[TIER_TWO] < _allowedLandSalePerChunk
```
poshdan, line 253: this stood out to me, is this a copy paste error, or is there a limit correlation between tier 2s and 3s?

bilbolPrime: I guess you are correct. The only constraints on T3 purchase is enough total t3 land is available and the chunk itself has not had its t3 land pruchased.

contracts/CBKLand.sol
bilbolPrime: chunkid is set as 0 when you purchase a T1 land because they haven't been distributed yet on the map. It's a random land but the algorithm to distribute them isn't decided yet. A round robin would end up with everything being too far apart.

contracts/Blacksmith.sol
```diff
@@ -301,4 +323,54 @@ contract Blacksmith is Initializable, AccessControlUpgradeable {
...
+        if(currency == 0){
+             game.payContractTokenOnly(payer, paying);
+        }
+        else if(currency == 1){
+            require(currencies[currency] != address(0), "invalid");
+            IERC20(currencies[currency]).transferFrom(payer, address(this), paying);
+        }
+    }
+
+    function recoverCurrency(uint256 currency, uint256 amount) public isAdmin {
```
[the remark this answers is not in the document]
bilbolPrime: Will remove, not sure why I missed it

migrations/91_cbkLandSale.js
bilbolPrime: I was going to ask to have conditional checks to know what address to put but that is not practical and would mean a migration for every new token. So I will add functions to add currencies.

PR #800, opened 2021-10-21
739-land-sale reserved sales

bilbolPrime:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

reserved sales

### DIFF
--- a/contracts/CBKLandSale.sol
+++ b/contracts/CBKLandSale.sol
@@ -276,8 +276,6 @@ contract CBKLandSale is Initializable, AccessControlUpgradeable {
         chunkT2LandSales[chunkId]++;
         chunkZoneLandSales[chunkIdToZoneId(chunkId)]++;
 
-        purchaseAddressMapping[buyer] = purchaseInfo(buyer, TIER_TWO, chunkId, false);
-
         emit T2GivenReserved(msg.sender, buyer, chunkId);
         cbkLand.mint(buyer, TIER_TWO, chunkId, uri);
     }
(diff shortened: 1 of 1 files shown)


contracts/CBKLand.sol
```diff
@@ -38,7 +41,7 @@ contract CBKLand is Initializable, ERC721Upgradeable, AccessControlUpgradeable {
     }
 
     function _restricted() internal view {
-        require(hasRole(GAME_ADMIN, msg.sender), "Not game admin");
+        require(hasRole(GAME_ADMIN, msg.sender) || hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "NA");
```
poshdan, line 44: is this only for the convenience of console admins? it's a nitpick over a principle, probably only like 100 gas so I'm not too bothered, just wondering

contracts/CBKLandSale.sol
```diff
@@ -358,25 +430,53 @@ contract CBKLandSale is Initializable, AccessControlUpgradeable {
         _enabled = allowed;
     }
 
-    function reserveChunks(uint256[] calldata chunkIds, bool reserve) external isAdmin {
+    function setReservedSaleAllowed(bool allowed) external isAdmin {
+        _reservedEnabled = allowed;
+    }
+
+    function reserveChunks(uint256[] calldata chunkIds, address reserveFor, bool reserve, bool forced) external isAdmin {
```
poshdan, line 437: a small note on the reserve parameter, it could be a bit confusing. But since this is an admin only function it's not a big deal. Ideally the function or param name would indicate the two directions this function can set things.

contracts/CBKLandSale.sol
```diff
@@ -217,16 +235,66 @@ contract CBKLandSale is Initializable, AccessControlUpgradeable {
...
+
+    function giveT2LandReserved(address buyer, uint256 chunkId, string memory uri) public reservedChunkAvailable(chunkId) reservedSaleAllowed() {
+        if(chunkT2LandSales[chunkId] == 0){
+            chunksWithT2Land++;
+        }
+
+        chunkT2LandSales[chunkId]++;
+        chunkZoneLandSales[chunkIdToZoneId(chunkId)]++;
+
+        purchaseAddressMapping[buyer] = purchaseInfo(buyer, TIER_TWO, chunkId, false);
```
poshdan, line 279: is this operation missing from the two other reserve functions? Or is this mistakenly present here and neither 3 are meant to have it

poshdan (changes requested):

contracts/CBKLandSale.sol
bilbolPrime: correct, not sure why I keep missing these. Removing.

contracts/CBKLandSale.sol
bilbolPrime: will rename to setChunksReservationInfo and rework a bit

PR #671, opened 2021-08-14
119 skill shop smoke bomb , exp scroll

bilbolPrime:
### All Submissions

* [ ] Can you post a screenshot of your changes (if applicable)?
* [ ] 
image
image
image
image

### New Feature Submissions

* [ ] Does this relate to an existing issue, or has it been talked about prior to being done (if a major change)?
* [ ] #119 

### Notes

Please note, your code must pass all tests and lint checks before it can be merged.

### PR Description

Unfortunately there was no avoiding in modifying the cryptoblade contract for the fetch target function. For the fight function, I kept the change inside the character contract inside the unpack method. I avoided modifying the structure of character and added a mapping for character seed. The seed can be used for other reasons in the future.

I made it 10x smoke bombs for 0.1 total.

Also added purchase event to blacksmith for easier purchase tracking.

EDIT: Added exp scroll, purchase events and some other fixes

Remaining: art for the 10x smoke bomb and 4x exp scroll

PR #824, opened 2021-10-31
119-smoke-bomb-v2 smokebomb skillshop item

bilbolPrime:
### All Submissions

* [ ] 
image
image
image
image

### New Feature Submissions

* [ ] #119 

### Notes

### PR Description

A redo of PR https://github.com/CryptoBlades/cryptoblades/pull/671 as it's ancient. Only did the smoke bomb part as the EXP scroll had some question marks on it. Also dropped the dynamic price part due to time constraints. They don't seem to be changing anyway.

~~I reworked the smoke bomb from chain side to avoid any change in gas cost to the fight function. I am 99.99% sure there is no change and I am seeing similar gas results but there is some variance even before the change.~~

Per @poshdan request, we are now just incrementing stamina by 1. This will make this have 0 effect on fight gas.

Made 50 smoke bombs cost only 0.1 SKILL. I think making them drop from combat is a good idea.

This needs to be put on testnet for at least 1 week. Straight forward but sensitive change.

Note: Use of smokebomb seems to be ~70k gas


frontend/src/store.ts
```diff
@@ -3166,7 +3166,51 @@ export function createStore(web3: Web3) {
...
         ]);
       },
+      async fetchTotalSmokeBombsOwned({ state }) {
+        const { SmokeBombConsumables } = state.contracts();
+        if(!SmokeBombConsumables || !state.defaultAccount) return;
+        return await SmokeBombConsumables.methods.getItemCount().call(defaultCallOptions(state));
+      },
+      async purchaseSmokeBomb50({ state, dispatch }) {
+        const { CryptoBlades, SkillToken, SmokeBombConsumables, Blacksmith } = state.contracts();
+        if(!CryptoBlades || !SmokeBombConsumables || !Blacksmith || !state.defaultAccount) return;
```
Chriserus, line 3176: Add SkillToken check here also

frontend/src/store.ts
```diff
@@ -3166,7 +3166,51 @@ export function createStore(web3: Web3) {
...
+
+        try {
+          await SkillToken.methods
+            .approve(CryptoBlades.options.address,  web3.utils.toWei('0.1', 'ether'))
```
Chriserus, line 3180: Minor suggestion, can't this price be passed by component using this method? Or we can use getSmokeBombPrice method, as it won't work if the price changes

frontend/src/store.ts
```diff
@@ -3166,7 +3166,51 @@ export function createStore(web3: Web3) {
...
+            .send({
+              from: state.defaultAccount
+            });
+        } catch(err) {
+          console.error(err);
+        }
+
+        await Blacksmith.methods.purchase50SmokeBombs( web3.utils.toWei('0.1', 'ether')).send({
```
Chriserus, line 3188: Same here with the price

frontend/src/views/Combat.vue
```diff
@@ -69,6 +69,13 @@
                 <b-button v-if="selectedWeaponId" variant="primary" class="ml-3" @click="selectedWeaponId = null" id="gtag-link-others" tagname="choose_weapon">
                   Choose New Weapon
                 </b-button>
+
+                <b-button v-if="selectedWeaponId" variant="primary" class="ml-3 smoke-bomb-button"
+                @click="useSmokeBombUI()" id="gtag-link-others" tagname="use_smokebomb"
```
Chriserus, line 74: @click="useSmokeBombUI" <- you can delete () here

frontend/src/views/Combat.vue
```diff
@@ -69,6 +69,13 @@
...
+                :disabled="smokeBombs == 0">
```
Chriserus, line 75: :disabled="smokeBombs ===  0"> 
Hmm, linter should suggest this change, do you have it set up correctly in your IDE? Also, as 0 is falsy, it can be just :disabled="!smokeBombs" (if smokeBombs is 0 it will be disabled, otherwise it will be enabled)

Chriserus (changes requested):

frontend/src/store.ts
kcper: Please use `approveFee` function to approve skill spend - this handles reading the price from chain.

kcper:
Contract code looks ok, just needs addressing Chriserus comments and resolve conflicts.

frontend/src/views/Combat.vue
bilbolPrime: done

frontend/src/store.ts
bilbolPrime: used approveFeeFromAnyContract but had to do a modification which allows sending payToAddress as a param. Fee contract in this case (blacksmith) isn't the pay to address (cryptoblades).
Tested minting and we should be fine

frontend/src/store.ts
bilbolPrime: done

frontend/src/store.ts
bilbolPrime: done

frontend/src/views/Combat.vue
bilbolPrime: for some reason it's only working as ==

bilbolPrime: It works when === and string ('0'); thanks for the tip

Chriserus (approved):

PR #922, opened 2021-12-15
Fix/pvp fixes

Jcalcaldev:


contracts/PvpArena.sol
```diff
@@ -57,6 +57,8 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
     uint8 public wageringFactor;
     /// @dev percentage of duel cost charged when rerolling opponent
     uint256 public reRollFeePercent;
+    /// @dev percentage of entry wager charged when withdrawing from arena with pending duel
+    uint256 public withdrawFeePercent;
```
poshdan, line 61: you cannot change the order of state variables. New ones have to be added after all the others

contracts/PvpArena.sol
```diff
@@ -201,6 +203,7 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
         // TODO: Tweak these values, they are placeholders
         wageringFactor = 3;
         reRollFeePercent = 25;
+        withdrawFeePercent = 25;
```
poshdan, line 206: just a heads up this is not going to be initialized on the testnets we've already deployed on

contracts/PvpArena.sol
```diff
@@ -370,7 +373,7 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
     }
 
     /// @dev performs all queued duels
-    function performDuels(uint256[] calldata attackerIDs) external restricted {
+    function performDuels(uint256[] memory attackerIDs) public restricted {
```
poshdan, line 376: this probably makes no difference but it's worth knowing that in some cases, external functions can handle arrays of data more efficiently (in terms of gas). An extra internal function might be ideal here considering that the restricted modifier is going to run (adding to gas) when called from the restartRankedSeason(), which is already checked against restriction. Not big enough of a deal to bother with this know, but it's worth mentioning.

poshdan (changes requested):

contracts/PvpArena.sol
Jcalcaldev: We added a setter for this value so you can call in the new migration file.

contracts/PvpArena.sol
Jcalcaldev: Noted!

Issue #887, opened 2021-12-02
Default alignment change of CharacterList


kevinunger:
How it is now:
![image](https://user-images.githubusercontent.com/5601589/144473487-805185d9-e8d5-41ef-a663-0d5145e949e3.png)

`justify-content:center`
![image](https://user-images.githubusercontent.com/5601589/144473531-c1a8404e-b01e-427c-b10e-3dafbf2e7d9e.png)

Bridge.vue already has centered chars by setting 
```
/deep/ .character-list{
  justify-content: center;
}
```
![image](https://user-images.githubusercontent.com/5601589/144473703-e8d1e836-4c27-42a3-8ffb-60a14c48b654.png)

Would be better to set the alignment to center (looks better imo) in CharacterList.vue. Weapons and other nfts are centered as well, so it would be consistent.
```
.character-list{
   ...
  justify-content: center;
}
```

kevinunger:
in market:
![image](https://user-images.githubusercontent.com/5601589/146359100-58795a73-e204-4c32-bf02-cf93d4f82ec6.png)
->
![image](https://user-images.githubusercontent.com/5601589/146359139-7748b6aa-9ab0-4632-999b-009e8b743a7a.png)

brian-man:
@seiyria Maybe float this on testnet and ask Early Testers for feedback?

seiyria:
for now, we'll avoid this change

PR #925, opened 2021-12-16
+center character-list

kevinunger:
- see issue #887


kcper:
@seiyria's call with this one I guess.

seiyria:
If this is desired then go for it. I'm not super sure personally. We can always do it and see what the reaction is to determine whether or not we should keep it.

Issue #910, opened 2021-12-07
Idle Characters


seiyria:
You should be able to have an "idle character" garrison. These characters would not be part of your team, and would not recover stamina. 

You should be able to buy characters from the market or recruit new characters, even if your team is full. These characters would go to your "garrison". You could also choose to swap someone out of your main party and push them into reserve when buying or recruiting a character (or just send them to reserve, the default action).

This is important for #492, and should be done beforehand.

brian-man:
@seiyria Anything remaining work needed to close this issue?

seiyria:
nope, this is done

PR #950, opened 2021-12-22
Garrison

kcper:
### All Submissions

image
image

### New Feature Submissions

#910 
### Notes

### PR Description

Front-end part:
Unlimited and seamless recruiting, receiving, purchasing and transferring to garrison. If any of these actions exceeds the 4 character limit, the character lands in garrison. Characters in garrison do not regenerate stamina.

Contract:
Override for safeTransferFrom function which redirects characters user's garrison if they would exceed the plaza limit - this is the simplest way to handle all scenarios/actions.
Added stamina lock/unlock functions to prevent regeneration in garrison.


contracts/Garrison.sol
```diff
@@ -0,0 +1,134 @@
...
+                tokens[i] = id;
+        }
+    }
+
+    function balanceOf(address user) public view returns(uint256) {
+        return userGarrison[user].length();
+    }
+
+    // MUTATIVE
+    function sendToGarrison(uint256 id) public {
```
Chriserus, line 72: Shouldn't there be some restrictions? Like checks, if the id provided is user's character? This method is public

frontend/src/contracts.ts
```diff
@@ -250,7 +254,7 @@ export async function setUpContracts(web3: Web3): Promise<Contracts> {
   }
 
   const pvpContracts: PvPContracts = {};
-  if(featureFlagPvP){
+  if(!featureFlagPvP){
```
Chriserus, line 257: Why is this change here? If PvP is on -> initialize contract, if not -> don't do it, it will crash the app. Or am I wrong here?

contracts/Garrison.sol
kcper: Not needed, it will just fail on safeTransferFrom if sender is not the owner of the nft. We could add a require statement just to throw an explicit error about what happened, but there's no way that happens when using our frontend for this call.

frontend/src/contracts.ts
kcper: Nice catch, I was playing with it earlier I guess.

contracts/characters.sol
```diff
@@ -116,6 +121,10 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
     mapping(uint256 => mapping(uint256 => uint256)) public nftVars;//KEYS: NFTID, VARID
     uint256 public constant NFTVAR_BUSY = 1; // value bitflags: 1 (pvp) | 2 (raid) | 4 (TBD)..
 
+    Garrison public garrison;
+    mapping(uint256 => bool) public isStaminaLocked;
+    mapping(uint256 => uint8) public lockedStaminaPoints;
```
bilbolPrime, line 126: This will start failing if we max stamina exceeds 255, perhaps we should make it uint16 just to be safe otherwise this will become a zombie state var in the future

contracts/characters.sol
```diff
@@ -198,8 +215,16 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
 
             tokens.push(Character(xp, level, trait, staminaTimestamp));
             cosmetics.push(CharacterCosmetics(0, RandomUtil.combineSeeds(seed, 1)));
-            _mint(minter, tokenID);
-            emit NewCharacter(tokenID, minter);
+            address receiver = minter;
+            if(minter != address(0) && minter != address(0x000000000000000000000000000000000000dEaD) && !hasRole(NO_OWNED_LIMIT, minter) && balanceOf(minter) >= characterLimit) {
+                receiver = address(garrison);
+                garrison.redirectToGarrison(minter, tokenID);
+                _mint(address(garrison), tokenID);
```
bilbolPrime, line 222: Not saying this is wrong but redirecting before the actual minting feels weird. No transfer is being made so this is fine but it's like we are emitting events and modifying state vars on something that needs one more line of code before "existing". Just being picky in case the redirectToGarrison in the future gets updated to actually transfer the mint. You can ignore this one if that will never happen.

bilbolPrime (changes requested):

contracts/characters.sol
kcper: Yeah I was trying to invoke it after transfer initially, but `onERC721Received` was causing the transaction to fail.
It does this
```
require(
    supportedTokenTypes.contains(_tokenAddress) &&
        allCharactersInGarrison.contains(_id),
    "Token ID not listed"
);
```
and `allCharactersInGarrison` are updated in `redirectToGarrison` function.
The thing is `onERC721Received` has to make use of `id` parameter and only `allCharactersInGarrison.contains(_id),` made sense to me.

bilbolPrime (approved):

PR #962, opened 2021-12-30
Feat/pvp overhaul  finish contract

Jcalcaldev:
We remade the PvPArena contract to implement the new opponent assignation strategy, plus multiple corrections and fixes on many functions.

This PR only involves the contract. Tests and frontend update coming soon.


contracts/PvpArena.sol
```diff
@@ -227,167 +233,232 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
...
-        weapons.setNftVar(weaponID, 1, 1);
+        excessWagerByCharacter[characterID] = 0;
+        fighter.wager = 0;
+
+        skillToken.safeTransfer(msg.sender, wager);
     }
 
-    /// @dev attempts to find an opponent for a character
-    function requestOpponent(uint256 characterID)
+    /// @dev attepts to find an opponent for a character
```
brian-man, line 317: attempts

contracts/PvpArena.sol
```diff
@@ -728,124 +688,214 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
...
+        uint256 opponentID;
+        uint256 matchableCharactersCount = matchableCharacters.length();
+        bool foundOpponent = false;
+
+        for (uint256 i = 0; i < matchableCharactersCount; i++) {
+            uint256 index = (randomIndex + 1) % matchableCharactersCount;
+            uint256 candidateID = matchableCharacters.at(index);
+
+            if (candidateID == characterID) {
+                if (matchableCharactersCount == 1) {
```
brian-man, line 809: This should be checked before the for loop rather than inside it.
Checking it in the loop is a waste of gas and the answer won't change while iterating.

contracts/PvpArena.sol
```diff
@@ -728,124 +688,214 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
...
+        uint256 randomIndex = RandomUtil.randomSeededMinMax(
+            0,
+            matchableCharacters.length() - 1,
+            seed
+        );
+        uint256 opponentID;
+        uint256 matchableCharactersCount = matchableCharacters.length();
+        bool foundOpponent = false;
+
+        for (uint256 i = 0; i < matchableCharactersCount; i++) {
```
brian-man, line 804: The loop index `i` is never used.

I suspect the intent was to use it here (replacing the 1):
`uint256 index = (randomIndex + 1) % matchableCharactersCount;`

contracts/PvpArena.sol
```diff
@@ -728,124 +688,214 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
...
+                if (
+                    matchableCharacters.at(matchableCharactersCount - 1) ==
+                    candidateID
+                ) {
+                    candidateID = matchableCharacters.at(0);
+                } else {
+                    candidateID = matchableCharacters.at(index + 1);
+                }
+            }
+            if (
```
brian-man, line 821: This logic appears to be preventing the characters in the same account from competing with each other in PVP.

Recommend:
`_matchableCharactersByTier` should change from:
`mapping(uint8 => EnumerableSet.UintSet) private _matchableCharactersByTier;`
to
`mapping(uint8 => mapping(address => EnumerableSet.UintSet)) private _matchableCharactersByTier;`

Then the for loop disappears entirely.  You pick a random index.  That index indicates an account with 1-4 characters ready to be matched.  If the index picks the current character, then use the next slot (with wraparound) which is guaranteed to be a different account.

Note that the explicit check `if (matchableCharactersCount == 1)` may also become unnecessary.

brian-man: Sorry, proposed data structure was wrong above.

```
struct MatchableAccount {
    address owner;
    EnumerableSet.UintSet characters;
}

mapping(uint8 => MatchableAccount) _matchableAccountsByTier;
```

Existing
- `uint256 candidateID = matchableCharacters.at(index);`
- check in the for loop to see if account is matching and continue loop if so

Proposed
- `MatchableAccount matchableAccount = matchableAccounts.at(index);`
- check `matchableAccount.owner` for match, if matched fetch `matchableAccount = matchableAccounts[(index+1)%matchableAccountsCount]`
- you can assert that the 2nd matchableAccount owner is different
- no loop

kcper (approved):

PR #978, opened 2022-01-04
Feat/pvp overhaul  frontend

RafaelVidaurre:
* Updates pvp frontend to adjust to contract changes

### DIFF
--- a/frontend/src/components/smart/PvPArenaPreparation.vue
+++ b/frontend/src/components/smart/PvPArenaPreparation.vue
@@ -1,6 +1,8 @@
 <template>
-    <div class="arenaPreparationWrapper">
-    <div class="mainWrapper">
+    <div v-if="loading">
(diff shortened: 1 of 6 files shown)


frontend/src/components/smart/PvPArenaPreparation.vue
```diff
@@ -295,6 +307,7 @@ export default {
         console.log(this.entryWager);
         console.log('Missing data');
```
Chriserus, line 308: Console logs, probably delete it

frontend/src/components/smart/PvPArenaPreparation.vue
```diff
@@ -295,6 +307,7 @@ export default {
...
       }
+      this.loading = false;
```
Chriserus, line 310: With all the loadings and such, if there is a try catch and after that some loading is set to false regardless of the outcome -> I would put it in finally. Maybe not this particular place ofc

Chriserus (approved):

PR #995, opened 2022-01-07
Feat/pvp  kick event

Jcalcaldev:
This PR:

Emits an event when a character is kicked from the arena due to not having enough wager to continue dueling.

Processes the event in the front end to inform players correctly when their character was kicked, removing the impression that sometimes your character just appears outside of the arena.


Chriserus (commented):
LGTM, I mainly checked frontend

Jcalcaldev:
@Chriserus Added missing translations, will get the missing ones soon as well.

Chriserus (approved):

PR #1000, opened 2022-01-10
resolves #793

### LINKED ISSUE #793: Automatic chain switch based on user's current metamask network
Currently the app tries to connect to the network that is selected in Options and switching between networks in metamask has no effect. We want the app to adjust the chain it tries to connect to (and set it in options) based on user's current metamask network.

Also right now options page is not accessible if user is on a wrong network. Make sure the options page will be always accessible (should be by default if the above is implemented).

kevinunger:
- resolves #793  - Automatic chain switch
- Added some lines in common.ts so that an event is triggered when the user changes network from Metamask
- localstorage is then filled with the chain name and the page reloads
- we could now change the 'You can't currently view the app right now. You can try clearing LocalStorage to reset your settings.' error to something else, because it only appears, when the user selects a testnet when on prod or vice versa. It's also shown briefly, after the page reload

### DIFF
--- a/frontend/src/utils/common.ts
+++ b/frontend/src/utils/common.ts
@@ -1,11 +1,34 @@
 import axios from 'axios';
 import BigNumber from 'bignumber.js';
 import Web3 from 'web3';
+import config from '../../app-config.json';
 
 BigNumber.config({ ROUNDING_MODE: BigNumber.ROUND_DOWN });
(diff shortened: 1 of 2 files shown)


Chriserus (approved):
lgtm

kevinunger:
added fix not to use color js >1.4.0

PR #1003, opened 2022-01-11
Fixed CPU issue relating to weapon lists

Sellie96:
Removed the infinite animation adding a glow to weapons. Was causing massive CPU loads and memory leaks.


Chriserus (approved):
lgtm

PR #1006, opened 2022-01-11
fix: duel modal displaying incorrect info and misc ui fixes

Jcalcaldev:
This PR:

Fixes inconsistencies in the duel won/lost modal.

Also has some small UI fixes.


Chriserus (approved):
LGTM

Issue #492, opened 2021-07-28
character burn / reforge


seiyria:
Characters will get a "rebirth level". You can buy characters from the market and burn them to buff one of your characters. This will require:

- [ ] a new "rebirth level" being tracked on the NFT. this will go up by the level of the character burnt. So if I burn a level 5 character onto my level 15 character, it will get a new stat for rebirth, set to 5. If not set, it will buff.
- [ ] solidity work to allow you to burn people from the market, and people in your party if so chosen (similar to character reforge)
- [ ] frontend work to allow you to burn a character instead of buy it, as well as to burn a character in your party to buff another one.
- [ ] solidity work to multiply the "rebirth level" by 3 when adding power to the characters combat score

This score will apply only to PvP and raids.

aquahamzah:
What is the decided max "rebirth level" we want this to go to? if i burn a level 51 e.g - that will provide 51 rebirth.

seiyria:
There is no max. The bonus value will be tuned based on what we see, but 3 is a good place to start.

poshdan:
Is this a "flat" power bonus or something that gets multiplied by weapons?
A character's starting stamina is enough to get it to level 9 almost, i think a lot of people will be trying to exhaust the stamina before burning, it would only cost them ~0.0041 BNB to do the fights, so it's almost a no brainer to get 9x the benefit.

seiyria:
@poshdan flat power bonus. I'm comfortable tiering it too, so `level / 10` is the boost. Then you only get +1 for burning a level 1-9, but +2 at 10.

poshdan:
Yeah the tiering could help, but i think a lot of peple will still get 10-11 for the double benefit, since its not far off 9.

Right now minting a character costs like $23, I could easily see some users investing several characters worth of burns, and whales might do up to a hundred, at least in the near future.
I presume the mechanic is intended as a sink primarily for low level characters, so i guess we don't have to do anything special on burning high levels, even though they take exponentially longer to level up.

A level 51 with an awesome 5 star sword and LB could be reaching 37k power,
A level 41 with good weapon and LB might be around 20k power.
An average joe with a level 21 and decent sword would be maybe 11k.

Should the level tier bonus still apply to this (lv 51 being 6x, so 3 power per burnpoint is actually 18)? I think it's a nice balance between letting a newcomer get a head start vs rewarding long term play.

We need to figure out some power percentages that would make this worthwhile for everyone.

[...]

PR #1015, opened 2022-01-13
Char burn

kcper:
### All Submissions

https://www.loom.com/share/089c5bcfda2b46039e02249d69b0b70b
### New Feature Submissions

#492 
### Notes

### PR Description

Character burning feature. Needs soul.png replaced with our desired art.


frontend/src/views/Market.vue
```diff
@@ -1565,7 +1593,7 @@ export default Vue.extend({
       this.allSearchResults = Array.from(this.allSearchResults as string[]).filter((x: any) => x.id !== this.selectedNftId);
 
       this.waitingMarketOutcome = false;
-      this.marketOutcome = 'Successfully purchased '
+      this.marketOutcome = 'Successfully purchased ' + (isBurning ? 'and burned ' : '')
```
Chriserus, line 1596: Missing translation I think

frontend/.env.local
```diff
@@ -9,3 +9,4 @@ VUE_APP_API_URL=""
 VUE_APP_API_KEY="accessToken"
 VUE_APP_FEATURE_FLAG_NFT_BRIDGE="1"
 VUE_APP_FEATURE_FLAG_MERCHANDISE="1"
+VUE_APP_FEATURE_FLAG_BURNING_MANAGER="1"
```
Chriserus, line 12: Are other flags set correctly? Please double check

frontend/src/store.ts
```diff
@@ -1461,6 +1471,14 @@ export function createStore(web3: Web3) {
         ]);
       },
 
+      async fetchSoulBalance({ state }) {
+        const { Characters } = state.contracts();
+        if(!Characters || !state.defaultAccount) return;
+        const soulBalance = await Characters.methods.soulSupply(state.defaultAccount).call(defaultCallOptions(state));
```
Chriserus, line 1477: Immediately return, your IDE should suggest that

frontend/src/views/Market.vue
```diff
@@ -700,6 +700,23 @@
         </div>
       </b-tab>
     </b-tabs>
+    <b-modal class="centered-modal text-center" ref="character-buy-modal" :title="$t('market.characterBuyModal')"
+      @ok="purchaseNft" @cancel="purchaseNft(true)">
```
Chriserus, line 704: On cancel call purchaseNft?

Chriserus (approved):

frontend/src/views/Market.vue
kcper: Yes, there's no cancel option, but I'm using the cancel handler event.

Chriserus (approved):

contracts/characters.sol
```diff
@@ -125,16 +125,20 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
 
     Garrison public garrison;
 
+    mapping(uint256 => uint256) public bonusPower;
```
poshdan, line 128: I'd really like for us to use the nftVars mapping for things like this.
In the long run, it will save on setters and getters, as well as allow us to place logic outside of the contract (like into the burning manager. We could even move soulSupply there.
Even the nft bridge will have an easier time grabbing bulk nftVars in an array rather than calling multiple getters.

contracts/characters.sol
```diff
@@ -26,7 +26,7 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
     }
 
     function migrateTo_1ee400a() public {
-        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Not admin");
+        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "NA");
```
poshdan, line 29: would recommend removing the "NA" strings entirely from all migration functions. Saves a decent bit on contract size

contracts/characters.sol
```diff
@@ -240,6 +244,46 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
...
+        _burn(burnId);
+        soulSupply[soulRecipient] = soulSupply[soulRecipient].add(soulAmount);
+
+        emit Burned(
+            soulRecipient,
+            burnId
+        );
+    }
+
+    function upgradeWithSoul(uint256 targetCharId, uint256 soulAmount) public {
```
poshdan, line 279: should probably be an ownerOf check done for this action

contracts/characters.sol
```diff
@@ -252,6 +296,10 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
         return getPowerAtLevel(tokens[id].level);
     }
 
+    function getTotalPower(uint256 id) public view noFreshLookup(id) returns (uint256) {
```
poshdan, line 299: if we stored total power in an nft var on levelup and soul upgrade, we could save on calculating power for fights and other stuff that uses it (this will tie in to the equipment system too)
It's also important to discuss how pvp might use power, in case they don't want soul burns to affect their power, we might need separate values stored in which case the "totalpower" might require renaming to something like "pvePower"

contracts/BurningManager.sol
```diff
@@ -0,0 +1,98 @@
...
+        }
+        return burnFee;
+    }
+
+    function burnCharacterFee(uint256 burnId) public view returns (uint256) {
+        return (game.vars(game.VAR_HOURLY_PAY_PER_FIGHT()) / game.vars(game.VAR_HOURLY_MAX_POWER_AVERAGE())) * 7 * characters.getTotalPower(burnId) * 60;
+    }
+
+    //FUNCTIONS
+
```
poshdan, line 69: this applies to all of the following burn functions:
it might make sense to always use arrays instead of having a separate function for burning one character at a time.
And might make sense to accept arrays on the characters contract end, and iterate there, because external calls can take ~4k gas each, meaning you pay 1 cent extra on BSC per character/external call. (if you do ownerof checks in 2 places and then call the burn action, that's 3 cents per character. It can add up. There's further garrison calls from inside characters too)
Since burn manager has more room to spare, it might make sense to add arrayed views for things like owner checks etc.

contracts/BurningManager.sol
```diff
@@ -0,0 +1,98 @@
...
+    // MODIFIERS
+
+    modifier isCharacterOwner(uint256 burnId) {
+        _isCharacterOwner(burnId);
+        _;
+    }
+
+    function _isCharacterOwner(uint256 burnId) internal view {
+        require(hasRole(BURNER_ROLE, msg.sender) || characters.ownerOf(burnId) == msg.sender || garrison.characterOwner(burnId) == msg.sender, 'Not owner');
```
poshdan, line 40: this role check might not make sense. This contract is the source of calls on the soul burns in characters (which are restricted), and the only thing that should have this role (afaik) is the market

poshdan (changes requested):
I want us to confirm what the needs will be for power in pvp and perhaps move logic to burnmanager and use arrays in functions

contracts/characters.sol
kcper: There's no way in the front-end to upgrade characters one doesn't own, but might be useful for some nerds to have that option available in the contract, why not?

contracts/BurningManager.sol
kcper: Only market has this role. What's a better way of checking if it's the market making the call?

poshdan: right, i forgot this was a logical AND statement. So not technically an issue. But it's certainly a loaded check when we're iterating over multiple characters.
I guess for now it's fine to leave it as is, unless a smart solution presents itself. Alternatively you can have an extra burn function that is just for the role.

contracts/characters.sol
```diff
@@ -244,44 +251,46 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
...
     }
 
-    function upgradeWithSoul(uint256 targetCharId, uint256 soulAmount) public {
-        require(soulSupply[msg.sender] >= soulAmount, 'Not enough soul');
+    function upgradeWithSoul(uint256 targetCharId, uint256 soulAmount) external restricted {
         uint256 burnPower = soulAmount.mul(10);
         require(uint(4).mul(getPowerAtLevel(tokens[targetCharId].level)) >= getTotalPower(targetCharId).add(burnPower), "Power limit");
-        soulSupply[msg.sender] = soulSupply[msg.sender].sub(soulAmount);
-        bonusPower[targetCharId] = burnPower.add(bonusPower[targetCharId]);
+        nftVars[NFTVAR_BONUS_POWER][targetCharId] = burnPower.add(nftVars[NFTVAR_BONUS_POWER][targetCharId]);
```
poshdan, line 293: The order of keys is flipped for nftVars here.
It's supposed to be targetCharId first, then nft var second

contracts/BurningManager.sol
```diff
@@ -31,23 +34,14 @@ contract BurningManager is Initializable, AccessControlUpgradeable {
...
-
     modifier isCharactersOwner(uint256[] memory burnIds) {
         _isCharactersOwner(burnIds);
         _;
     }
 
     function _isCharactersOwner(uint256[] memory burnIds) internal view {
         for(uint i = 0; i < burnIds.length; i++) {
-            require(hasRole(BURNER_ROLE, msg.sender) || characters.ownerOf(burnIds[i]) == msg.sender || garrison.characterOwner(burnIds[i]) == msg.sender, 'Not owner');
+            require(characters.ownerOf(burnIds[i]) == msg.sender || garrison.characterOwner(burnIds[i]) == msg.sender, 'Not owner');
```
poshdan, line 44: in the future we will be able to add a function to characters and garrison that does this check in bulk, for now it's okay unless we find gas costs unreasonable to begin with

contracts/characters.sol
poshdan: actually in a couple other places too (line 309, line 271, 257)

poshdan (changes requested):
looks mostly good now I think, just a minor oopsie to fix

poshdan (approved):

PR #1023, opened 2022-01-14
Feat/pvp  handle season restart

Jcalcaldev:
This PR:

- Adds a counter so players know when a PvP ranked season restarts.
- Displays current season number.
- Adds an indicator on the "Rewards" tab so players know when they can claim rewards from being top players in a season.
- Notifies players the instant a season is restarted.

### DIFF
--- a/frontend/src/components/smart/PvPArena.vue
+++ b/frontend/src/components/smart/PvPArena.vue
@@ -91,6 +91,7 @@ export default {
       seasonStartedAt: null,
       seasonDuration: null,
       secondsBeforeNextSeason: null,
+      restartEventSubscription: null,
       characterInformation: {
         tier: null,
         name: '',
@@ -290,7 +291,7 @@ export default {
       let blockToScan = initialBlock;
       let scanning = false;
 
-      this.web3.eth.subscribe('newBlockHeaders', async (_, result) => {
+      const subscription = this.web3.eth.subscribe('newBlockHeaders', async (_, result) => {
         try {
           if (scanning) {
             return;
@@ -324,6 +325,8 @@ export default {
           scanning = false;
         }
       });
+
+      this.restartEventSubscription = subscription;
     }
   },
 
@@ -438,8 +441,10 @@ export default {
     await this.updateSecondsBeforeSeason();
   },
 
-  unmounted() {
-    this.web3.eth.clearSubscriptions();
+  destroyed() {
+    if (this.restartEventSubscription.id) {
+      this.restartEventSubscription.unsubscribe();
+    }
   },
 
   watch: {
--- a/frontend/src/components/smart/PvPArena.vue
+++ b/frontend/src/components/smart/PvPArena.vue
@@ -14,6 +14,8 @@
         :availableShieldIds="availableShieldIds"
         :ownedWeaponsWithInformation="ownedWeaponsWithInformation"
         :ownedShieldsWithInformation="ownedShieldsWithInformation"
+        :currentRankedSeason="currentRankedSeason"
+        :secondsBeforeNextSeason="secondsBeforeNextSeason"
         @enteredArena="handleEnteredArena"
       />
       <pvp-arena-summary
@@ -24,6 +26,8 @@
         :activeWeaponWithInformation="activeWeaponWithInformation"
         :activeShieldWithInformation="activeShieldWithInformation"
         :duelHistory="duelHistory"
+        :currentRankedSeason="currentRankedSeason"
+        :secondsBeforeNextSeason="secondsBeforeNextSeason"
         @enterMatchMaking="handleEnterMatchMaking"
       />
       <!-- Should use router -->
@@ -83,6 +87,10 @@ export default {
       isMatchMaking: false,
       tierRewardsPool: null,
       tierTopRankers: [],
+      currentRankedSeason: null,
+      seasonStartedAt: null,
+      seasonDuration: null,
+      secondsBeforeNextSeason: null,
       characterInformation: {
         tier: null,
         name: '',
@@ -156,6 +164,7 @@ export default {
           information: await this.getShieldInformation(fighter.shieldID)
         };
       }
+
       this.loading = false;
     },
 
@@ -168,6 +177,13 @@ export default {
       this.isCharacterInArena = false;
       this.isMatchMaking = false;
       this.$emit('leaveMatchMaking');
+    },
+
+    async updateSecondsBeforeSeason() {
+      this.currentRankedSeason = await this.contracts().PvpArena.methods.currentRankedSeason().call({ from: this.defaultAccount });
+      this.seasonStartedAt = await this.contracts().PvpArena.methods.seasonStartedAt().call({ from: this.defaultAccount });
+      this.seasonDuration = await this.contracts().PvpArena.methods.seasonDuration().call({ from: this.defaultAccount });
+      this.secondsBeforeNextSeason = (+this.seasonStartedAt + +this.seasonDuration) - (Math.floor(Date.now() / 1000));
     },
 
     async updateOpponentInformation(defenderId) {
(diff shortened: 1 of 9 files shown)


frontend/src/components/smart/PvPArena.vue
```diff
@@ -170,6 +180,13 @@ export default {
       this.$emit('leaveMatchMaking');
     },
 
+    async updateSecondsBeforeSeason() {
+      this.currentRankedSeason = await this.contracts().PvpArena.methods.currentRankedSeason().call({ from: this.defaultAccount });
```
Chriserus, line 184: Don't use direct contract calls in vue files, keep it in store.ts, it applies to every call here

frontend/src/components/smart/PvPArena.vue
```diff
@@ -372,10 +437,22 @@ export default {
     this.loading = false;
   },
 
+  async updated() {
+    await this.updateSecondsBeforeSeason();
+  },
+
+  destroyed() {
+    if (this.restartEventSubscription.id) {
```
Chriserus, line 445: Can restartEventSubscription be null here? It will throw an exception if it's possible

frontend/src/components/smart/PvPArenaPreparation.vue
```diff
@@ -165,7 +165,18 @@
             <span>{{ tierTopRankers[2] && tierTopRankers[2].rank || 'N/A'}}</span>
           </li>
         </ul>
-        <!-- <a href="#" class="rankings">View all rankings</a> -->
+        <ul class="topPlayersList">
+          <li class="header">
+            <span>Current Season</span><span>Restarts In</span>
```
Chriserus, line 170: After it's live someone will come back and clean all the translations! I'm watching...

frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -72,11 +72,20 @@
...
           </li>
         </ul>
-        <!-- <a href="/" class="rankings">View all rankings</a> -->
+        <ul class="topPlayersList">
+          <li class="header">
+            <span>Current Season</span><span>Restarts In</span>
+          </li>
+          <li>
+            <span>{{ currentRankedSeason }}</span>
+            <vue-countdown :time="secondsBeforeNextSeason * 1000" v-slot="{ days, hours, minutes, seconds }">
```
Chriserus, line 81: I see a duplication here, with stuff from PvPArenaPreparation, can we make a component out of it? (everything in < ul >)

frontend/src/components/smart/PvPNavBar.vue
```diff
@@ -11,7 +11,13 @@
       </button>
       <button @click="setTab(2)" :class="tabNumber === 2 && 'active'">
         <img src="../../assets/crownIcon.svg" />
-        <span>Rewards</span>
+        <span>Rewards
+          <div class="rewardsIndicator" v-if="hasRewards">
+            <div>
+              <p>!</p>
```
Chriserus, line 17: I don't know how it will look like in browser, but I think in this case we can use an icon indicator and not literal "!"

frontend/src/views/PvP.vue
```diff
@@ -34,12 +34,13 @@ export default {
...
       tab: 0,
-      isCharacterMatchMaking: false
+      isCharacterMatchMaking: false,
+      hasRewards: false,
     };
   },
 
   computed: {
-    ...mapState(['currentCharacterId']),
+    ...mapState(['currentCharacterId', 'contracts', 'defaultAccount']),
```
Chriserus, line 43: Don't. Please extract to store.ts

Chriserus (changes requested):

frontend/src/components/smart/PvPArena.vue
RafaelVidaurre: @Chriserus I disagree with this, passing all calls through the store has no real value if no app state is affected. The main issue with the previous implementation of this feature was that it's whole state was passed through the store even though it was specific to this view.

Also that meant we had to manually reset the state on route changes.

frontend/src/components/smart/PvPArenaPreparation.vue
Jcalcaldev: Will be done

frontend/src/components/smart/PvPArenaInfo.vue
```diff
@@ -0,0 +1,172 @@
...
+.title {
+  margin-bottom: 1rem;
+  color: #cec198;
+  font-size: 1.25rem;
+  font-family: 'Trajan';
+  line-height: 1.75rem;
+  padding: 0;
+}
+.tokenCard {
+display: flex;
```
Chriserus, line 100: No indent

frontend/src/components/smart/PvPArenaInfo.vue
```diff
@@ -0,0 +1,172 @@
...
+  .number {
+    color: #ffffff;
+    font-size: 1.25rem;
+    line-height: 1.75rem;
+  }
+}
+}
+.topPlayersList,
+.characterAttrsList {
+display: flex;
```
Chriserus, line 129: Same here. Fixing this should be automatically done by IDE if configured

frontend/src/components/smart/PvPArena.vue
Chriserus: "no real value", it's basic single responsibility principle. It's the same with any api calls, why do we create services for that? Won't it work directly from component? Ofc it will, but will be non-extendable and the SRP will be broken. That's why we have layers (components call api through services, web3 through store.ts) and at least some sort of architecture in our project. I know that single, huge store.ts is not ideal, I'm aware of that, but it's better to continue extracting calls there then to litter components with a responsibility that shouldn't be handled there.

Also, the solution you guys provided is error prone because it's not checking contracts or defaultAccount existence and they're not always available (see checks in every method in store).

Happy to discuss it further.

Chriserus: Also, "that meant we had to manually reset the state on route changes", I'm not talking about handling state of PvP in store. I just mean that when you have a simple getter like the line we're commenting under, it should be called this way:

this.currentRankedSeason = await this.currentRankedSeason(); 
In which, currentRankedSeason is an action imported from the store (ideally with an interface here showing what are the arguments and what is the output)

Chriserus: Example interface from another class: 
```
interface StoreMappedActions {
  fetchShields(shieldIds: (string | number)[]): Promise<void>;
  fetchWeapons(weaponIds: (string | number)[]): Promise<void>;
  fetchCharacters(characterIds: (string | number)[]): Promise<void>;
}
```

RafaelVidaurre: Replying to this now, sorry for the delay.

About SRP: I think this is not a case of SRP violation, one store doing every web3 call has multiple responsibilities. I think there's clear value in extracting reusable code, but these calls are literally only used for this component. I do think there's a strong case here in terms of this PR being inconsistent with the rest of the codebase though.

About "Checking for defaultAccount existence (should be)". I guess we assumed this values were guaranteed to be set by the time the route rendered. If not then yup, we have to do that, though I'd say we should consider this in the future as part of an architecture improvement, we shouldn't be repeating this logic on every view if we can extract it.

I think moving the calls themselves to the store is OK, if that's how we are doing it everywhere else, but the previous implementation had horrible issues with the **state** being global. If your request is just about moving the web3 calls into actions then we can definitely do that.

TL;DR;

* Happy to move web3 calls to actions for consistency
* I'm against moving view-specific state to the store (though I think I misunderstood, and you weren't suggesting this)
* I have to dissagree with this being a SRP issue in this feature, and I'd even argue it *is* a big issue on the store itself, as it should be subdivided into modules that do respect that principle (right now being at 4k+ lines), so if anything, that's where we should be looking to make improvements regarding separation of concerns first.

One last thing though. I think we should merge this PR and tackle the migration of web3 calls to actions. Because there's already multiple instances of these merged, and I think it is better if we tackle this on a separate PR

Chriserus: I'm aware of our overall architectural issues with having just one store, we're on the same page here. I didn't want you to move any pvp-specific states to the store, that's ok.

Just for the sake of consistency, please move all the web3 calls to store.ts, so we have it all in one place and not scattered in components.

I'm happy with doing it on a different PR in the future, I just want you to make sure it won't get lost due to new features or something, cleanup should take place in not-so-distant future.

As my other comments were formatting-related I will approve the PR and you can merge it in appropriate time.

@seiyria tagging you here, as we had a nice discussion with Rafael and I want it to be visible

Chriserus (approved):

frontend/src/components/smart/PvPArena.vue
RafaelVidaurre: Thanks for the help @Chriserus , will do that

PR #1033, opened 2022-01-17
Feat/pvp  add under attack modal

Jcalcaldev:
This PR:

Changes the old generic alert for a custom modal to show players they're under attack.

### DIFF
--- a/frontend/src/components/smart/PvPUnderAttackModal.vue
+++ b/frontend/src/components/smart/PvPUnderAttackModal.vue
@@ -1,0 +1,196 @@
+<template>
+  <div @click="close" class="modalWrapper">
+    <div class="modalInnerWrapper">
+      <img src="../../assets/separatorTop.svg" alt="Top separator">
+      <div class="modalTitle">You are currently under attack!</div>
+      <ul>
+        <li>
+          <span>
+            Please wait a moment
+          </span>
+        </li>
+      </ul>
+    </div>
+    <div class="closeWrapper">
+      <p>{{$t('pvp.tapAnywhere')}}</p>
+      <button @click="close">
+        <img src="../../assets/closeModal.svg" alt="Bottom separator">
+      </button>
+    </div>
+  </div>
+</template>
+
+<script>
+
+export default {
+  props: {
+    isUnderAttack: {
+      default: null
+    }
+  },
+
+  data() {
+    return {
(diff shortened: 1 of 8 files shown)


frontend/src/components/smart/PvPUnderAttackModal.vue
```diff
@@ -0,0 +1,196 @@
+<template>
+  <div @click="close" class="modalWrapper">
```
Chriserus, line 2: Why are you using a custom modal and not b-modal (like everywhere else)? Is this needed?

frontend/src/locales/fr.json
```diff
@@ -584,7 +584,9 @@
   "pvp": {
     "lastDefeated": "Vous avez été vaincu pour la dernière fois par :",
     "tapAnywhere": "Appuyez n'importe où pour fermer",
-    "kicked": "Vous avez été expulsé de l'arène !"
+    "kicked": "Vous avez été expulsé de l'arène !",
```
Chriserus, line 587: Just so you know, you don't need to provide all the translations :D There's always fallback for english, and translations can be handled by translators. Nice addition nevertheless!

Chriserus (approved):

frontend/src/components/smart/PvPUnderAttackModal.vue
Jcalcaldev: Not absolutely, we can write an issue to migrate our custom modals to b-modals if you wish so

frontend/src/locales/fr.json
Jcalcaldev: Too late! Got most of the translations ready in another PR, should I delete them and leave them to the translators or leave them in? I would assume some of them might be wrong due to Google translate, but I don't know what's easier for them :o

Chriserus: Just leave it however you've done it :D Not a problem for them, just to assure you, you don't need to provide ALL translations, just english, that's why I was surprised you guys didn't want to do them on the spot. It's just adding a key value in en.json, the rest will be handled.

frontend/src/components/smart/PvPUnderAttackModal.vue
Chriserus: Please do

PR #1040, opened 2022-01-19
resolves #1031

### LINKED ISSUE #1031: query params to load specific chains
supported chains can be seen [here](https://github.com/CryptoBlades/cryptoblades/blob/main/frontend/app-config.json) - "OEC", "BSC", "HECO", "POLYGON", "AVAX"

when a user is on a chain, if their chain was in the url as `?chain=<chain>`, that would work for sharing. when loading a page and the query param has `?chain=<chain>`, it would be ideal to force the user onto that chain. if unspecified, it should load their current/default chain, then set the query param for the future.

kevinunger:
- issue #1031 
- Added query params 

Changes in App.vue in router watcher:
- Check if params are in URL, if so check if chain is supported --> change network in MM and set localstorage 'currentChain'
- If chain in params === chain in localStorage or no params given --> add params to URL

Changes in Options.vue:
- change of chain from the options page --> add chain to params

Changes in Common.ts:
- if change of chain comes from MM --> add chain to params

Used router.replace the way I did to avoid a refresh and to avoid the redundant navigation warning from vue-router

### DIFF
--- a/frontend/src/views/Options.vue
+++ b/frontend/src/views/Options.vue
@@ -303,6 +303,12 @@ export default Vue.extend({
     async setCurrentChain() {
       localStorage.setItem('currentChain', this.currentChain);
       Events.$emit('setting:currentChain', { value: this.currentChain });
+      (this as any).$router.replace(
+        {
+          query: Object.assign({ ...(this as any).$route.query }, { chain: this.currentChain }),
+        },
+        () => {}
+      );
       await this.configureMetaMask(+getConfigValue('VUE_APP_NETWORK_ID'));
     },
   },
--- a/frontend/src/views/Options.vue
+++ b/frontend/src/views/Options.vue
@@ -101,6 +101,7 @@ import i18n from '../i18n';
 import { getConfigValue } from '@/contracts';
 import config from '../../app-config.json';
 import { SupportedProject } from './Treasury.vue';
+import { addChainToRouter } from '@/utils/common';
 
 interface StoreMappedState {
   skillRewards: string;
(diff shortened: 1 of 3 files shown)


kevinunger:
and  @Chriserus maybe you know how to replace the (this as any).$router [here](https://github.com/CryptoBlades/cryptoblades/pull/1040/commits/83dbbcba27afcf04c9fbd6e073165493ff2dca3b#diff-d38c5bc7527876792e5ddab4c3306cf37b17036f8e94d7e8bf8d0545e783b3c7R303-R314)
Typescript error when not doing the (this as any)-thing is this:
```Property '$router' does not exist on type 'CombinedVueInstance<Vue, Data, { toggleGraphics(): void; toggleRewards(): void; toggleAdvanced(): void; onClaimTokens(): Promise<void>; claimSkill(stage: ClaimStage): Promise<...>; ... 11 more ...; fetchPartnerProjects(): Promise<...>; }, { ...; }, Readonly<...>>'.```
:man_shrugging:

https://github.com/vuejs/vue/issues/8721

**Comment by kevinunger** [comment]
If anyone wants to test locally (on production or testnet (beware, testnet avax/polygon/oec are reaaaly slow)) change:
https://github.com/CryptoBlades/cryptoblades/blob/83dbbcba27afcf04c9fbd6e073165493ff2dca3b/frontend/src/utils/common.ts#L24 to `const env = 'production' `
and in contracts.ts:
(https://github.com/CryptoBlades/cryptoblades/blob/83dbbcba27afcf04c9fbd6e073165493ff2dca3b/frontend/src/contracts.ts#L95-L96)
comment L95 
L96: `const env='production'`

File: frontend/src/views/Options.vue [hunk 1]
```diff
@@ -303,6 +303,12 @@ export default Vue.extend({
     async setCurrentChain() {
       localStorage.setItem('currentChain', this.currentChain);
       Events.$emit('setting:currentChain', { value: this.currentChain });
+      (this as any).$router.replace(
```
kcper, line 306: There's a router being used in common.ts without `as any` conversion. You could create and export a function in common.ts that will do `router.replace` and call it from here.

Chriserus: Checked it, should work, you can also change (this as any).$router that's in NftDisplay component

kevinunger: @kcper @Chriserus Thanks, worked!

kcper (approved):

PR #1041, opened 2022-01-19
Updated styling on avg stat

Sellie96:
image

Updated styling to reduce the size, prominence and truncate to 2 decimal places


kevinunger (approved):

kcper (approved):
Nice

PR #1045, opened 2022-01-19
use correct router instance

kevinunger:
common.ts was creating it's own router instance so that router.replace could not add to params to current route, because the current route was always "/".


Chriserus (approved):
lgtm

kcper (approved):

PR #1087, opened 2022-01-25
Feat/pvp  optimization

Jcalcaldev:
This PR:

-Optimizes gas costs in the PvpArena contract.
  * This comes with extra mantainability costs, such as the contract calculating isTraitEffectiveAgainst for itself instead of calling another contract's function.
  * In local unit tests, gas cost were reduced from 376074 to 355378.

-Fixes previously broken tests.
-Fixes an issue with weaponId props in the front end.


brian-man:
@Jcalcaldev
> This comes with extra mantainability costs
Put the functions into a common .sol file that both contracts can use share.

PR #1089, opened 2022-01-26
Gas optimization: PvpArena.performDuels() uses getTrait() directly

brian-man:
Estimated gas savings: 10k gas per shield (so 0-20k gas per duel)


Chriserus (commented):
LGTM, but need another reviewer to take a look

Jcalcaldev (approved):
LGTM

PR #1092, opened 2022-01-27
Safe Randoms

poshdan:
Extensive notes on usage and details available in the SafeRandoms.sol file,
But to summarize:

- Safe randoms offer seeds that are secure (provided you follow guidelines)
- Users help us resolve previous seeds (optionally), so we don't _need_ to run a service
- Seeds cannot be resolved during the same transaction or block they are requested at, requiring two transactions to process the result
- There are ways to chain calls to effectively make it 1 action per transaction, but the very first one will likely require 2 anyway
- **"Safe" is a relative term - It depends on network activity**. The more features that use it, the better. Resolving a seed 1 block after the request is practically secure.
- If you use the resolve step on pop functions, you will resolve your seed immediately (not having to wait for others), IF it was requested on a previous block

Comes with a mint N function for weapons, something we will want to leverage as soon as possible with the use of Safe Randoms.

**Gas figures:** (from local network)
(includes external contract step costs (5k) except for the resolve step)
(doesn't include base transaction cost (21k if chained, 42k if done strictly 2-tx))
(doesn't include event costs (3k each) or "requestNext" costs)

- Resolving a seed costs about ~35k gas
- Requesting a random seed costs ~30k gas for the single variant (~31k to salt first time, 26k to salt consecutive times), ~50k for queued variant (~35k for consecutive ones)
- Popping a single seed costs ~1k gas, popping a queued seed costs ~7k for consecutive ones, ~3k to pop the last (first queued) one


kcper (approved):
I had it explained to me like I'm five and now I can say it looks good.

Chriserus (approved):
I also had that explained to me like I'm in diapers even, looks good to me (and I needed this kind of explanation! :D )
I know that one more change is coming, lmk when it's pushed @poshdan

poshdan:
Done.
Would like eyes on
https://github.com/CryptoBlades/cryptoblades/pull/1092/commits/39dce7ae840a7c49d4998c6b6c4c538715af675b
to tell me i'm not crazy
@Chriserus @kcper @bilbolPrime

kcper:
> Done. Would like eyes on [39dce7a](https://github.com/CryptoBlades/cryptoblades/commit/39dce7ae840a7c49d4998c6b6c4c538715af675b) to tell me i'm not crazy @Chriserus @kcper @bilbolPrime

It looks good to me.

bilbolPrime (approved):

PR #1106, opened 2022-01-31
Add getPowerAtLevel() to common.sol

brian-man:
Expected to save 7310 gas per duel in `performDuels()`.

### DIFF
--- a/contracts/common.sol
+++ b/contracts/common.sol
@@ -6,6 +6,7 @@ import "abdk-libraries-solidity/ABDKMath64x64.sol";
 library Common {
     using ABDKMath64x64 for int128;
     using SafeMath for uint256;
+    using SafeMath for uint8;
 
     function isTraitEffectiveAgainst(uint8 attacker, uint8 defender) internal pure returns (bool) {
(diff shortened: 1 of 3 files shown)


contracts/common.sol
```diff
@@ -18,4 +19,20 @@ library Common {
     ) internal pure returns(uint24) {
         return uint24(weaponMultiplier.mulu(basePower).add(bonusPower));
     }
-}
\ No newline at end of file
+
+    function getPowerAtLevel(uint8 level) internal pure returns (uint24) {
+        return Common.getPowerAtLevel(level);
```
poshdan, line 24: recursive call. I think you copied the code from characters

brian-man: Yah, typo.  Will remove the cut/paste snafu.

poshdan (approved):
Looks good now.

In the future, deleting the public getPowerAtLevel function from Characters would be a way to save a bit on the contract size. I have a hunch the frontend might be using it but we could duplicate the calculation for the frontend (not to mention skipping the delay on the numbers)

PR #1107, opened 2022-01-31
_getShieldStats() no longer calls characters.getTrait()

brian-man:
Expected to save 5669 gas per shield per duel in `performDuels()`.


poshdan:
Still a relevant optimization but this is in the wrong contract (pr slipped through the cracks earlier, sorry. PvpCore is the relevant file now)
I would suggest opening a new pr, possibly with some other changes merged in, as I see the structs aren't carried all the way currently, it could benefit from more saved calls to character level etc.

PR #1108, opened 2022-01-31
Add getDefenseMultiplierForTrait() variant to shields.sol

brian-man:
External calls to `shields.getFightData()` cost 15479 gas.  PvpArena only uses one of the four return values.

This change adds a `getDefenseMultiplierForTrait()` variant so that PvpArena can call it directly.


poshdan (approved):
I want to point out that the gas difference between the two functions is 2k. Two calls per duel, I guess it's something.
This can be slotted in the next time we're upgrading pvparena/shields

PR #1120, opened 2022-02-01
Feat/pvp  ranked rewards on burnt chars

Jcalcaldev:
This PR:

-Fixes an issue with restartRankedSeason if one of the top 3 players leaves the arena and burns it's character.
-Updates shield minting in tests (broken since last week).
-Should add some gas saving in some functions by calling _ownerByCharacter instead of characters.ownerOf


poshdan:
Shouldn't isOwnedCharacter also use the new _ownerByCharacter value?

Jcalcaldev:
> Shouldn't isOwnedCharacter also use the new _ownerByCharacter value?

This is optimal but would cause some characters to be trapped in an invalid state in current ongoing networks (like BSC Testnet). Since _ownerByCharacter is set upon entering arena, players who have their characters already inside would not be able to leave and re-enter. I'm not sure how to approach this issue in the most optimal way.

poshdan:
It should be possible to expel players via an admin call.
The only concern there is gas, but on OEC it's very negligible.
We shouldn't be making compromises when a little elbow grease can help bridge the gap

contracts/PvpArena.sol
```diff
@@ -118,6 +118,8 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
     uint256 public duelOffsetCost;
     address payable public pvpBotAddress;
 
+    /// @dev owner's address by character ID
+    mapping(uint256 => address) private _ownerByCharacter;
     mapping(uint256 => uint256) public specialWeaponRerollTimestamp;
```
poshdan, line 123: the new mapping must come after the last state variable, please correct

poshdan (changes requested):
one small change needed afaik

kcper (approved):

poshdan (approved):

PR #1109, opened 2022-02-01
added shields for bridging

kevinunger:
- issue #1080 pr #1081 

-Shields are now shown all 3 tabs 
image
image
-Updated the third tab to be more responsive
image
image

- added isBridge prop in NftList.vue that disables filter options

- added css clamps for mobile

-To Do (couldn't find a quick way to split the string for nfts != weapons||chars without breaking stuff; maybe its too late)
https://github.com/CryptoBlades/cryptoblades/blob/37e581d5281daeacf2b01685903f5bfd17dd4f55/frontend/src/views/Bridge.vue#L446-L453
(split the <type>.<id> to --> <id> then do the checks)

Issue #1080, opened 2022-01-25
Bridge shields


bilbolPrime:
Bridge shields

poshdan:
they done did it!

PR #1121, opened 2022-02-01
issue-1080-additional-changes

kevinunger:
- issue #1080 #1081
- refractor
- resolved what @Chriserus said in PR #1109
  - added types/interfaces for bridge in store from(./interfaces/nft.ts)
    - transferedNft (for an nft that has been transferred)
    - nftTransfer (active transfer object)
  - changed the @click 
  - fixed the canWithdraw bug by first checking if `id` (if its weapon/char)  or `type.id` can be split ( removing the `type` entirely => items in nft-list can't be selected)
   - this.selectedNft is kept as String, and parsed in the methods for the above reason
- Moved the 'select target chain' - input select lower (was confusing - even for me when it was on top of the 'current transfer', bc it has nothing to do with it)
- fix typo in en.json


frontend/src/interfaces/Nft.ts
```diff
@@ -23,6 +23,26 @@ export type NftType = 'weapon' | 'character' | 'shield';
 
 export const allNftTypes: NftType[] = ['weapon', 'character', 'shield'];
 
+export interface transferedNft {
```
Chriserus, line 26: A really minor thing, interfaces should be kept in Pascal case, so, TransferedNft and NftTransfer :)

frontend/src/views/Bridge.vue
```diff
@@ -373,19 +375,19 @@ export default Vue.extend({
...
       targetChain: '',
-      targetChainId: '',
+      targetChainId: 0 as number | null,
       supportedChains: [] as string[],
-      supportedChainIds: [] as string[],
+      supportedChainIds: [] as number[],
       transferStatus: '',
       transferStates,
-      currentTransferNFTId: '',
+      currentTransferNFTId: 0 as number | null,
```
Chriserus, line 383: We should usually use `undefined` as a "no value", this way, in interfaces, when you declare the field that can be either a number or a "no value", you can use
`currentTransferNFTId?: number`, this way it's optional.

But it's nothing to worry about, just a tip

frontend/src/views/Bridge.vue
```diff
@@ -583,56 +596,58 @@ export default Vue.extend({
...
-
-      this.currentTransferChain = this.supportedChains[this.supportedChainIds.indexOf(currentTransferChainId)];
+      this.currentTransferNFTId = transfer.nftId;
+      this.currentTransferChain = this.supportedChains[this.supportedChainIds.indexOf(transfer.chainId)];
 
+      const currentTransferTokenAddress: string = transfer.nftAddress;
       if(currentTransferTokenAddress === this.Weapons.options.address) this.currentTransferNFTType = 'weapon';
       else if (currentTransferTokenAddress === this.Characters.options.address) this.currentTransferNFTType = 'character';
       else this.currentTransferNFTType = 'shield';
+      console.table(transfer);
```
Chriserus, line 650: Console.table to delete I guess :D

Chriserus (approved):

PR #1124, opened 2022-02-02
Feat/pvp  shield factor

Jcalcaldev:
This PR:

-Adds a shieldFactor variable to regulates shield's influence on PvP.
-shieldFactor must be set upon deployment.


contracts/PvpArena.sol
```diff
@@ -1010,7 +1017,7 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
 
         int128 bonusShieldStats;
         if (fighter.useShield) {
-            bonusShieldStats = _getShieldStats(character.ID);
+            bonusShieldStats = _getShieldStats(character.ID).mul(int128(_shieldFactor)).div(100);
```
poshdan, line 1020: i don't think the mul function is correct here.
ABDKMath64x64.sol says
/**
   * Calculate x * y rounding down.  Revert on overflow.
   *
   * @param x signed 64.64-bit fixed point number
   * @param y signed 64.64-bit fixed point number
   * @return signed 64.64-bit fixed point number
   */
  function mul (int128 x, int128 y)

What you probably want to use is .mulu(

  /**
   * Calculate x * y rounding down, where x is signed 64.64 fixed point number
   * and y is unsigned 256-bit integer number.  Revert on overflow.
   *
   * @param x signed 64.64 fixed point number
   * @param y unsigned 256-bit integer number
   * @return unsigned 256-bit integer number
   */
  function mulu (int128 x, uint256 y)

poshdan (changes requested):

poshdan (approved):
looks ok now. I'd like to see it through testnet before main

PR #1125, opened 2022-02-02
nft-list should only show shields on bridge

kevinunger:
only display shields when using nft-list on bridge

### DIFF
--- a/frontend/src/components/smart/NftList.vue
+++ b/frontend/src/components/smart/NftList.vue
@@ -611,6 +611,10 @@ export default Vue.extend({
         }
       }
 
+      if(this.isBridge){
+        return this.shieldsWithIds(this.nftsToDisplay.map(x => x.id.toString())).filter(Boolean);
+      }
+
       if(this.isReward && this.showGivenNftIdTypes) {
         const rewardedDust = this.nftsToDisplay.filter(x => x.type?.startsWith('dust')).map(x => { return { type: x.type, id: 0, amount: x.amount }; });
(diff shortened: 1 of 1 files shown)


kcper (commented):
Isn't this the problem?
Birdge.vue line 58
```
<div class="sell-grid" v-if="nftType === 'shield'">
          <nft-list v-model="selectedNftId" :isBridge="true"
          />
</div>
```
In other places withing Bridge.vue there's this prop passed in `<nft-list />`:
`:nftIdTypes="[currentTransferNFTId].map(id => { return { id: id, type: 'shield' }; })"`

This mechanism should limit shown nfts to the types you specify in `nftIdTypes`.

kevinunger:
Wouldn't that require getting the owned shields in bridge and passing them to nft-list? Isn't it easier to only return shields from nft-list.vue, when isBridge==true, like it's done with isMarket?

kcper:
> Wouldn't that require getting the owned shields in bridge and passing them to nft-list? Isn't it easier to only return shields from nft-list.vue, when isBridge==true, like it's done with isMarket?

`ownedShieldIds` are already stored in state so there's no problem with that.
Your solution makes sense too, but the naming doesn't. I'd either rename `isBridge` to `showShields` or something along these lines or even better in my opinion if you'd do a prop like `nftType` and filter `nftsToDisplay` (here `return this.nftsWithIdType(this.nftsToDisplay).filter(Boolean);`) with the type from `nftType` prop.

PR #1128, opened 2022-02-04
Oracle gas improvement for price conversion

brian-man:
The following contracts have at least one price oracle member and use the oracle `currentPrice()` method whenever doing conversions for at least 5,000 gas per call:
- Blacksmith.sol
- cryptoblades.sol
- NFTMarket

The following contracts invoke the oracle `currentPrice()` method via the game contract (cryptoblades.sol) at a cost of at least 10,000 gas per call:
- PvpArena.sol
- raid1.sol

This change introduces registration to the `IPriceOracle` (using a new `IPriceListener` interface) to allow consumers of pricing be notified ("push") of price updates so that they no longer have to query ("pull") to obtain the pricing.

With this change none of the above contracts invoke `currentPrice()` and thus player-invoked methods no longer pay gas to obtain the current pricing.

The caller of `setCurrentPrice()` on oracles now pays a bit more to update all of the registered listeners to provide this service to them.

### DIFF
--- a/contracts/BasicPriceOracle.sol
+++ b/contracts/BasicPriceOracle.sol
@@ -3,12 +3,15 @@ pragma solidity ^0.6.5;
 import "@openzeppelin/contracts-upgradeable/proxy/Initializable.sol";
 import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
 import "./interfaces/IPriceOracle.sol";
+import "./interfaces/IPriceListener.sol";
 
 contract BasicPriceOracle is IPriceOracle, Initializable, AccessControlUpgradeable {
     bytes32 public constant PRICE_UPDATER = keccak256("PRICE_UPDATER");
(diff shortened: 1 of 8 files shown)


poshdan:
**Please don't do this.
We didn't ask for this, it's not a good use of your time, or ours.**

Only the pvp contract had a significant gas issue we wanted to tackle, but you went on to bloat several other contracts with this band-aid assuming it's worth the hassle.
The old oracle hasn't been updated for months and at the moment it's unclear what we will do about automatic price adjustment in the future (if anything).
Blacksmith.sol only uses the oracle for the lands sales, which actually uses external deployments of the oracle which we'd have to explicitly upgrade and set up, and it only ever reports to one place.

Even if we were to operate the old oracle again regularly, now there's several contracts it has to delegate calls to every time the price is updated - now you've exchanged 5 or at most 10k gas cost on the user for 40-50k gas added to the update function, which would triple the operating cost of the oracle. (or more if we kept adding listeners)

You are welcome to submit the pvp arena changes in another PR with the usd variable added, and a local setter.

PR #1150, opened 2022-02-08
Feat/pvp  leave arena button

Jcalcaldev:
This PR:
- Adds a "leave arena" button in the arena summary. This allows players to leave arena without having to awkwardly enter match making to do so


frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -8,13 +8,22 @@
...
+              @click="handleEnterArenaClick"
+              :buttonText="$t('pvp.enterArena')"
+            />
+          </div>
+          <div class="buttonWrapperSecondary">
+            <pvp-button
+              @click="leaveArena"
+              :disabled="loading"
+              :buttonText="$t('pvp.leaveArena')"
+              :secondary="true"
```
Chriserus, line 24: Just so you know, if you have boolean props you can just add it without anything iirc
so, instead of `:secondary="true"` you should be able to use `secondary`

frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -135,7 +145,10 @@ export default {
   data() {
     return {
       tab: 0,
-      dayjs
+      dayjs,
+      loading: true,
```
Chriserus, line 149: By default loadings often are initialized to `false`, then on the action start, switched to `true`, after the action (often in `finally` block for actions that may fail) switched back to `false` again

frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -144,14 +157,61 @@ export default {
...
+
+      try {
+        await this.withdrawFromArena(this.currentCharacterId);
+
+        this.$emit('leaveArena');
+      } catch (err) {
+        console.log('leave arena error: ', err.message);
+
+        this.handleErrorMessage(err.message, 'Char not in arena', 'The character is not in the arena');
+        this.handleErrorMessage(err.message, 'Defender duel in process', 'Duel already in process');
```
Chriserus, line 197: Error messages are not translated, is there a reason for that? same on line 173

frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -144,14 +157,61 @@ export default {
...
+      }
+
+      this.loading = false;
```
Chriserus, line 200: I would place it in `finally` block, just for the sake of keeping it tidy, also, not sure if this gets called on try fail

frontend/src/components/smart/PvPArenaSummary.vue
```diff
@@ -144,14 +157,61 @@ export default {
...
+  async created() {
+    this.loading = true;
+
+    this.duelQueue = await this.getDuelQueue();
+
+    if (this.duelQueue.includes(`${this.currentCharacterId}`)) {
+      this.isCharacterInDuelQueue = true;
+    }
+
+    this.loading = false;
```
Chriserus, line 213: Same as before. In this function there's no `try...catch...finally` block, please add

Chriserus: If that's what needed for the given scenario, sure thing, go for it. Overall just being stuck on the "loading" action is incorrect, when you need to try something again, let users... try again :D But I don't know the whole context behind it, for sure it will make users refresh the page

Chriserus:
LGTM now, feel free to merge

PR #1153, opened 2022-02-08
Lower gas usage by characters._beforeTokenTransfer()

brian-man:
When invoking `characters.safeTransferFrom()` the `_beforeTokenTransfer()` function is invoked as a callback.  To set the promo bit, there are 4 external calls to promos.  This change replaces 4 promos external calls with 1.

In addition to direct usage by users, this method is invoked by the following:
- `Garrison.sendToGarrison()`
- `Garrison.restoreFromGarrison()`
- `NFTMarket.addListing()`
- `NFTMarket.cancelListing()`
- `NFTMarket.purchaseListing()`
- `NFTStorage.storeItem()`
- `NFTStorage.withdrawFromStorage()`

### DIFF
--- a/contracts/characters.sol
+++ b/contracts/characters.sol
@@ -18,6 +18,9 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
 
     bytes32 public constant MINTER_ROLE = keccak256("MINTER_ROLE");
 
+    // Copied from promos.sol, to avoid paying 5k gas to query a constant.
+    uint256 private constant BIT_FIRST_CHARACTER = 1;
+
     function initialize () public initializer {
         __ERC721_init("CryptoBlades character", "CBC");
         __AccessControl_init_unchained();
@@ -468,9 +471,10 @@ contract Characters is Initializable, ERC721Upgradeable, AccessControlUpgradeabl
 
     function _beforeTokenTransfer(address from, address to, uint256 tokenId) internal override {
         require(nftVars[tokenId][NFTVAR_BUSY] == 0);
-
-        promos.setBit(from, promos.BIT_FIRST_CHARACTER());
-        promos.setBit(to, promos.BIT_FIRST_CHARACTER());
+        address[] memory users = new address[](2);
+        users[0] = from;
+        users[1] = to;
+        promos.setBits(users, BIT_FIRST_CHARACTER);
     }
 
     function safeTransferFrom(address from, address to, uint256 tokenId) override public {


brian-man:
Sample polygon transaction:
https://polygonscan.com/tx/0xfaad3deb23c94e2a1d0827436da5e1b776cf17c4b5de70842a7d3143fbbd21be
> Gas Used by Transaction: 149,717

Trace:
https://polygonscan.com/vmtrace?txhash=0xfaad3deb23c94e2a1d0827436da5e1b776cf17c4b5de70842a7d3143fbbd21be&type=gethtrace2

Contract calls
```
DELEGATECALL 0x42842e0e safeTransferFrom() 130028/209438
    STATICCALL 0xbf2a28cf BIT_FIRST_CHARACTER() 7571/183673
        DELEGATECALL 0xbf2a28cf BIT_FIRST_CHARACTER() 243/173630
    CALL 0x260730e0 setBit() 5998/175766
        DELEGATECALL 0x260730e0 setBit() 5167/172240
    STATICCALL 0xbf2a28cf BIT_FIRST_CHARACTER() 1071/169379
        DELEGATECALL 0xbf2a28cf BIT_FIRST_CHARACTER() 243/165958
    CALL 0x260730e0 setBit() 3998/167876
        DELEGATECALL 0x260730e0 setBit() 3167/164473
```

Before
7571+1071 spent on 2x `BIT_FIRST_CHARACTER()`
5998+3998 spent on `setBit()`

After
~7000 spent on `setBits()`

Estimated savings
~11638
~7.7%

poshdan (approved):
I had a hunch this function would be run for mints and burns. I was right. I will be following up on weapons with a small tweak, because we could really use some cuts on mints and burns, however small.

I tested this, on my end it saved 10312 gas so that's more or less what you wrote (i separated the function from the nft flow, just calculating the beforeTransfer call in itself)
It also reduced the contract size by 0.16kb which surprised me a bit, considering we swapped a repeat function call to an array initialization. Turns out the new constant doesn't add enough to the contract size (due to it being private) to make even 0.01kb loss.

I'm approving this change. (reminder on the merge to add an upgrade migration)

What would be even better later on, if we could delete the whole promos bit altogether.

    function _beforeTokenTransfer(address from, address to, uint256 tokenId) internal override {
        if(from != address(0)) // this check saves like 300 gas on minting so why not
            require(nftVars[tokenId][NFTVAR_BUSY] == 0);
    }

This saves a further 0.27kb and reduces the gas cost of the transfers by 26472 (-16160 from this pr) gas if values were already set for both wallets, or 41472 (-31160 from this pr) gas if one wallet hasn't had it set (minting or transferring to a new alt)
(the case of both being unset is extremely uncommon (first ever mint etc))

I'll ask management if we can let go of this check

PR #1174, opened 2022-02-10
Feat/pvp  boosted roll

Jcalcaldev:
This PR:
- Adds a chance for attacker's to get a "boosted" roll in PvP duels.
- This chance will only happen if the opponent's weapons has more stars than the attacker's weapon.
- Winning a duel this way will give the attacker more MMR than usual, depending on a variable and the weapons' stars difference.
- Also adds some tests for this feature.

* boostChancePercentage and boostRankingPointsMultiplier must be set upon deployment.


poshdan:
If I understand correctly, this is designed to give the underdog a fighting chance by randomly favoring them in unbalanced fights in a gambling fashion.
And I believe this is made in response to high-end equipment being too powerful, discouraging lower players from pvp-ing at all.
From what I can tell there's some gas impact, though we've accepted that pvp won't be very economical on expensive chains.


Here's what I'm thinking:

Though the chance of the boost sounds exciting, it's important for the players to know this is happening - I have to express the necessity of frontend changes to communicate this to the player. (an event might even be necessary) Their gut reaction is not to fight a tough opponent unless we can convince them with some kind of data. If the duel screen said something like "20% chance to boost your roll against this opponent by 15%", that would go a long way.

The kicker here, is that the mechanic was abstracted without a clear idea of what numbers we'd want to use, and how they'll affect the average outcome. Correct me if I'm wrong here, I might have missed the discussion. We really need to have a ballpark on the specifics before committing to something like this.

Overall, I think we'll have a tough time balancing this. It does have a passive gambling appeal, but I would personally be in favor of a simpler solution.

poshdan (approved):
leaving my approval in case this gets pushed forward.
I still want to follow up with kroge on whether this is the way to go

poshdan:
Closing for now. We decided not to include the roll boost at this time (instead, ranking points are boosted from winning unfair fights).
We can revisit this later if need be

PR #1189, opened 2022-02-22
Feat/pvp  base100 formula update

Dvidaurre:
** add PlusMinus30Seeded() to utils.sol
** add getPlayerPowerBase100() formula to commons.sol
** adapt pvpArena to use it.

@poshdan Just let me know if you want me to create migrations on the PRs or if I should leave it to you


contracts/PvpArena.sol
```diff
@@ -1009,7 +1010,8 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
 
         int128 bonusShieldStats;
         if (fighter.useShield) {
-            bonusShieldStats = _getShieldStats(character.ID);
+            // we set bonus shield stats as 0.2
+            bonusShieldStats = _getShieldStats(character.ID).mul(20).div(100);
```
poshdan, line 1014: These weapon/shield multiplier values are 1-based, meaning the stats will add onto 1, not 0.
Since we will be adding the shield multiplier on top of the weapon, we need to counteract the base value of 1 to only add the stats, otherwise shields will still be too powerful (even a 1star shield with a single stat of 4 would give you +101% on your base character power).
The .sub(1) call would have to be done before the mul(20) step.

This is a very small change so I can handle it right before deployment

poshdan (approved):
looks good overall. One small thing I noted that I can do myself before deploying to testnet

contracts/PvpArena.sol
Dvidaurre: Perfect, understood Dan. Thanks for the correction!

PR #1191, opened 2022-02-24
Special weapons

kcper:
### All Submissions

Initial version here https://www.loom.com/share/c954218c2bcf4e17bd63f23292c72b7f 
Later on general shards have been removed and staking rewards added.
### New Feature Submissions

### Notes

### PR Description

(insert description here)

Introduces special forge for special weapons.
Covers requirements from https://docs.google.com/document/d/1hC9kQpv958X3tl-3jyIYdDLXSt42A-2qDx-hjCADXJo/edit and discussions from special-weapons channel.

### Testing

Everything apart from pvp free reroll has been tested locally.


poshdan:
I wanted to throw a curveball at the approach here.
What if we used the Promos contract for some of the event handling?
I don't know if the shard thing will be a permanent feature etc. And we may not want all the special weapon stuff inside the promos contract itself.
BUT, Promos is imported in a lot of places, it allows us to check for bad users, and the name fits the purpose.
It would probably add an external call in a number of places, which is of course sub optimal. But it may be cleaner, considering that we could save on migration functions.

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+      <b-tabs justified>
+        <b-tab :title="$t('blacksmith.forge')">
+          <div class="d-flex justify-content-center mt-3" v-if="isLoading">
+            <h4>{{$t('blacksmith.loading')}} <img class="spinner" src="../../assets/loadingSpinner.svg" /></h4>
+          </div>
+          <div v-if="!isLoading">
+            <div class="d-flex mt-3 align-items-center">
+              <h5 class="mb-0">{{$t('blacksmith.specialEvent')}}:</h5>
+              <b-form-select :disabled="!activeSpecialWeaponEventsIds.length" class="w-50 ml-1" size="sm"
+                v-model="selectedSpecialWeaponEventId" :value="selectedSpecialWeaponEventId" @change="updateSpecialWeaponEventId($event)">
```
Chriserus, line 12: Do you need :value here when you're already using v-model? Probably not

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+              <div class="d-flex flex-column align-items-center mt-3 top-border-forge" v-if="isEventWeaponOrdered && !isForging && !isEventWeaponForged">
+                <h5 class="mt-2">{{$t('blacksmith.specialWeaponOrdered')}}</h5>
+                <b-button
+                        variant="primary"
+                        @click="onClaimSpecialWeapon()"
+                        :disabled="isClaiming">
+                  <span v-if="!isClaiming" class="gtag-link-others">
+                    {{$t('blacksmith.claim')}}
+                  </span>
+                  <span v-if="isClaiming" class="gtag-link-others">
```
Chriserus, line 107: Can use v-else here, or v-else-if

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+                </div>
+              </div>
+            </div>
+          </div>
+        </b-tab>
+        <b-tab :title="$t('blacksmith.shards')">
+          <div class="d-flex justify-content-center mt-3" v-if="isLoading">
+            <h4>{{$t('blacksmith.loading')}} <img class="spinner" src="../../assets/loadingSpinner.svg" /></h4>
+          </div>
+          <div v-if="!isLoading">
```
Chriserus, line 125: v-else-if, or v-else

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+              </div>
+              <div class="d-flex justify-content-center mt-3">
+                <b-button
+                        variant="primary"
+                        @click="onConvertShards()"
+                        :disabled="!canConvert || isConverting">
+                  <span v-if="!isConverting" class="gtag-link-others">
+                    {{$t('blacksmith.convert')}} ({{convertAmount}} {{$t('blacksmith.shards')}})
+                  </span>
+                  <span v-if="isConverting" class="gtag-link-others">
```
Chriserus, line 179: v-else/v-else-if

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+                    <h5 v-if="availableElement === 100 || availableElement === 2" class="lightning-icon mb-0" />
+                    <h5 v-if="availableElement === 100 || availableElement === 3" class="water-icon mb-0 mr-2" />
+                  </div>
+                </div>
+                <div class="w-50 mb-3">
+                  <span>{{$t('blacksmith.endsIn')}}: {{eventEndsIn}}</span>
+                  <h4>{{partnerName}}</h4>
+                  <div class="mt-2">
+                    <h5 class="text-justify">{{eventDetails}}</h5>
+                    <a class="a-button" @click="openPartnerWebsite">{{eventPartnerWebsite}}</a>
```
Chriserus, line 36: :href="eventPartnerWebsite" target="_blank", no need for @click here

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -0,0 +1,641 @@
...
+</script>
+
+<style scoped>
+.weapon-img {
+  max-height: 250px;
+  max-width: max-content;
+  align-self: center;
+}
+.info-div {
+  padding: 2px;
```
Chriserus, line 604: Try to use p-1 or p-2 class for such padding, if possible

frontend/src/locales/en.json
```diff
@@ -288,7 +288,39 @@
...
     "couldNotBurn": "Could not burn sword: Insufficient funds or transaction was denied.",
-    "useMaxSoulAmount": "Use max"
+    "useMaxSoulAmount": "Use max",
+    "specialEvent": "Special Weapon Event",
+    "specialForge": "Special Forge",
+    "specialForgeTooltip": "Forge limited edition weapons created in partnership with other projects.",
+    "noActiveEvents": "There are no active events at this time.",
+    "pickSpecialEvent": "Pick an event to accumulate shards for",
+    "shards": "Shards",
+    "for": "for",
```
Chriserus, line 298: I'm not sure where it's used, but it's a better practice to just pass a number to the translation instead of cutting it, as it may differ in other languages

frontend/src/store.ts
```diff
@@ -5130,7 +5190,156 @@ export function createStore(web3: Web3) {
...
+
+          commit('updateSpecialWeaponEventsInfo', { eventId, eventInfo });
+        });
+      },
+
+      async fetchForgeCosts({ state }) {
+        const { SpecialWeaponsManager } = state.contracts();
+        if(!SpecialWeaponsManager || !state.defaultAccount) return;
+
+        const shardsCostLow = await SpecialWeaponsManager.methods.vars(1).call(defaultCallOptions(state));
```
Chriserus, line 5237: Doesn't matter that much, because the vars won't probably change the values, but you can use constants here (from the contract)

frontend/src/store.ts
```diff
@@ -5130,7 +5190,156 @@ export function createStore(web3: Web3) {
...
+        await SpecialWeaponsManager.methods.convertShards(eventFromId, eventToId, amount).send({ from: state.defaultAccount });
+
+        await dispatch('fetchShardsSupply');
+      },
+
+      async fetchShardsConvertDenominator({ state }) {
+        const { SpecialWeaponsManager } = state.contracts();
+        if(!SpecialWeaponsManager || !state.defaultAccount) return;
+
+        return +await SpecialWeaponsManager.methods.vars(10).call(defaultCallOptions(state));
```
Chriserus, line 5260: Same as above, if you feel like it, use constant from contract

frontend/src/views/Blacksmith.vue
```diff
@@ -34,6 +34,17 @@
...
                   {{$t('blacksmith.createDust')}}
                 </b-button>
+                <b-button
+                        variant="primary"
+                        class="ml-3"
+                        @click="onClickSpecialForge()"
+                        :disabled="disableForge"
+                        v-tooltip="$t('blacksmith.specialForgeTooltip')">
+                  <span v-if="disableForge">{{$t('blacksmith.coolingForge')}}</span>
+                  <span v-if="!disableForge" class="gtag-link-others">
```
Chriserus, line 44: v-else/v-else-if

frontend/src/views/Blacksmith.vue
```diff
@@ -580,28 +610,34 @@ export default Vue.extend({
...
-    const reforgeDustCost = await this.contracts.CryptoBlades.methods.reforgeWeaponWithDustFee().call({ from: this.defaultAccount });
-    const skillDustReforgeCost = await this.contracts.CryptoBlades.methods.usdToSkill(reforgeDustCost).call({ from: this.defaultAccount });
+    const reforgeDustCost = await this.contracts.BurningManager.methods.reforgeWeaponWithDustFee().call({ from: this.defaultAccount });
+    const skillDustReforgeCost = await this.contracts.BurningManager.methods.usdToSkill(reforgeDustCost).call({ from: this.defaultAccount });
     this.dustReforgeCost = new BN(skillDustReforgeCost).div(new BN(10).pow(18)).toFixed(4);
 
-    const burnCost = await this.contracts.CryptoBlades.methods.burnWeaponFee().call({ from: this.defaultAccount });
-    const skillBurnCost = await this.contracts.CryptoBlades.methods.usdToSkill(burnCost).call({ from: this.defaultAccount });
+    const burnCost = await this.contracts.BurningManager.methods.burnWeaponFee().call({ from: this.defaultAccount });
+    const skillBurnCost = await this.contracts.BurningManager.methods.usdToSkill(burnCost).call({ from: this.defaultAccount });
```
Chriserus, line 622: Web3 calls in component :x, I know you didn't do it, just swapped the contract, if you want, extract it to store, if not, leave as-is

contracts/BurningManager.sol
```diff
@@ -126,4 +237,21 @@ contract BurningManager is Initializable, AccessControlUpgradeable {
...
     }
+
+    function setBurnWeaponValue(uint256 cents) public restricted {
+        burnWeaponFee = ABDKMath64x64.divu(cents, 100);
+    }
+
+    function setReforgeWeaponValue(uint256 cents) public restricted {
+        int128 newReforgeWeaponFee = ABDKMath64x64.divu(cents, 100);
+        require(newReforgeWeaponFee > burnWeaponFee);
+        reforgeWeaponWithDustFee = newReforgeWeaponFee - burnWeaponFee;
```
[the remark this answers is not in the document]
kcper: I don't understand your comment + it was just copy pasted from weapons.sol as weapons contract cleanup was part of this PR.

frontend/src/store.ts
kcper: I wanted to avoid extra loading time.

frontend/src/views/Blacksmith.vue
kcper: There are so many instances of similar usage that I'd rather refactor it in a dedicated issue instead of doing little bits randomly and hoping we'll get rid of all of them eventually.

frontend/src/locales/en.json
kcper: It's not used in a sentence, just in dropdown options e.g. 3-5* `for` 10 Shards.

frontend/src/components/smart/SpecialWeaponForgeModal.vue
```diff
@@ -438,7 +439,7 @@ export default Vue.extend({
 
     forgedWeapon(): IWeapon | undefined {
       return this.weaponsWithIds(this.ownedWeaponIds).find(w => {
-        return w.weaponType === this.selectedSpecialWeaponEventId;
+        return w && w.weaponType === this.selectedSpecialWeaponEventId;
```
Chriserus, line 442: w && w.weaponType
is the same as
w?.weaponType

Chriserus (approved):

poshdan:
> > I wanted to throw a curveball at the approach here. What if we used the Promos contract for some of the event handling? I don't know if the shard thing will be a permanent feature etc. And we may not want all the special weapon stuff inside the promos contract itself. BUT, Promos is imported in a lot of places, it allows us to check for bad users, and the name fits the purpose. It would probably add an external call in a number of places, which is of course sub optimal. But it may be cleaner, considering that we could save on migration functions.
> 
> What would fit into "some" of the event handling? My initial approach was tracking 1 per wallet limit in Promos, but I quickly dropped this idea due to avoidable external calls. But the rest of the logic doesn't quite fit Promos contract in my opinion.
> 
> I'm pretty sure it would save on exactly one migration function in game contract while adding new migration function in Promos contract to add reference to a contract with the rest of special weapons logic (since we don't want all of it in Promos).
> 
> If we care to check for bad users in scope of special weapons minting we can add Promos contract to SpecialWeaponsManager and use it there.
> 
> Btw regarding the extra space these migration functions take up I think we could replace all existing ones in each contract with just one per contract, combine the logic and update migration files accordingly. We could keep doing that from time to time to save up on contract sizes.

I figured we can forward calls from promos to specialweapons. But it's not ideal for gas. So fair enough.

Something to consider as an alternative for keeping things clean, and this may not be applicable for this PR but for the future, we can utilize a "links" mapping (uint => address) that can store contract addresses (heck we can store it in vars and just convert from/to address on interaction), which allows us to hook in semi-temporary contracts in places that we need to expand sparingly. There is a small gas cost of the conversion process, but it's at most a few hundred iirc.

Raid1 has the links implementation for example. It allows us to import the abi, convert the address to the contract and call functions without creating a state variable. This also saves on the migration function aspect in the future, where we could pass an array of addresses and indices to store them all at once without having additional parameters.

Anyway. I haven't had the time to look through everything in detail. At the surface it looks good, I can trust Chriserus to have covered anything important.

PR #1205, opened 2022-03-02
Feat/pvp  tweaks

Jcalcaldev:
This PR:
- Adds a function in PvpArena.sol to get a character's full power (includes weapon's and shield's influence on the equation, excludes trait bonus).
- Displays full power in the front-end when inside the arena (weapon/shield already assigned to character).
- Multiplies trait bonus by 2, since PvP power roll volatility was increased.
- Fixes duel history not showing.
- Shortens some require-messages to reduce contract size.


kcper (commented):
Looks good to me

poshdan (approved):

PR #1209, opened 2022-03-04
Add discard support to NFTMarket

brian-man:
As discussed in Discord, this change adds support for automatically removing listed items when and address is marked banned.

To support discarding historical items and high volume addresses, intermediate `slice` functions are made available as well.

While some of these function are reasonable for an end user to use on owned accounts, all methods are currently `restricted` for now.

Note:
With this change the Character contract size is 23.98KB.  If we want, we can defer the support for characters in `discardListing()` until other changes have a chance to shrink the contract to give more breathing room.

Testing:
- Compiles and deploys locally.
- add admin and minion accounts to metamask
- build and deploy contracts
- metamask: transfer 100 skill from admin to minion
- game UI: recruit a character on minion
- game UI: forge weapon x10 on minion
- game UI: forge weapon x10 on minion
- game UI: fight once on minion (workaround for `noFreshLookup`)
- python script: minion invokes `NFTMarket.addListing()` for all 1 star weapons
- python script: verify weapons on market via `NFTMarket.getListingIDs(<Weapons>)`
- python script: admin invokes `NFTMarket.discardListing(<Characters>, 100)`
  => `Revert reason: Token ID not listed`
- python script: admin grants `GAME_ADMIN` role for self to Weapons
- python script: admin grants `GAME_ADMIN` role for NFTMarket to Weapons
- python script: admin invokes `NFTMarket.discardListing(<Weapons>, 20)`
  => completed
  => confirm `DiscardedListing` event is sent and contains correct parameters
  => confirm weapon no longer on NFT Market
  => confirm `Weapons.ownerOf(20)` throws an exception (no owner after burned)
- python script: admin invokes `NFTMarket.setUserBan(<minion>, true)`
  => completed
  => confirm NFTMarket is now empty of weapons

### DIFF
--- a/contracts/NFTMarket.sol
+++ b/contracts/NFTMarket.sol
@@ -645,6 +645,7 @@ contract NFTMarket is
         while(typeCount > 0) {
             IERC721 tokenAddress = IERC721(listedTokenTypes.at(typeCount - 1));
             discardedCount += discardSellerListingsOfToken(_seller, tokenAddress);
+            --typeCount;
         }
     }
 
--- a/contracts/NFTMarket.sol
+++ b/contracts/NFTMarket.sol
@@ -146,6 +146,11 @@ contract NFTMarket is
         IERC721 indexed nftAddress,
         uint256 indexed nftID
     );
+    event DiscardedListing(
+        address indexed seller,
+        IERC721 indexed nftAddress,
+        uint256 indexed nftID
+    );
     event PurchasedListing(
         address indexed buyer,
         address seller,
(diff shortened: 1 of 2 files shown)


kcper (commented):
Looks good to me.

poshdan:
Closing for now.
The ban function runs the discard step automatically, which could fail if there are too many for the maximum gas limit.
Yeah the slice function exists but it creates additional busy work when all we want to do is to stop a malicious actor dead in its tracks.
Personally I see the discard action as unnecessary, since the NFTs are already on the market wallet, if the ban holds, nobody has access to them, they are almost as good as burned. We can revisit later if there is a change of heart / specific need for this.

PR #1210, opened 2022-03-04
Increase memory for 'npm run build:frontend'

brian-man:
More than half of the most recent `npm run build:frontend` atempts have failed due to out of memory issues.

This change increases the memory setting from 512MB to 8GB which seems to be sufficient.


brian-man:
You can see the chain of failures (without the fix) here:
https://github.com/CryptoBlades/cryptoblades/actions/workflows/lint.yml

PR #1212, opened 2022-03-04
Fix compiler warnings in contracts

brian-man:
This change fixes all compiler warnings reported by contract compilation.

No functional changes.  Validated by compilation.

### DIFF
--- a/contracts/CBKLandSale.sol
+++ b/contracts/CBKLandSale.sol
@@ -745,18 +745,18 @@ contract CBKLandSale is Initializable, AccessControlUpgradeable {
             require(reservedTier == TIER_TWO || reservedTier == TIER_THREE, "NA");
 
             uint256 counter = reservedChunksCounter[reseller];
-            for(uint256 i = counter; i < counter + rcLength; i++) {
-                uint256 cId = reservedChunks[reseller].at(uint256(i % rcLength));
+            for(uint256 j = counter; j < counter + rcLength; j++) {
+                uint256 cId = reservedChunks[reseller].at(uint256(j % rcLength));
                 if(reservedTier == TIER_TWO) { // T2, find a spot with enough population
                     if(chunkT2LandSales[cId] < _allowedLandSalePerChunk) {
                         assignedChunkid = cId;
-                        reservedChunksCounter[reseller] = uint256(i + 1);
+                        reservedChunksCounter[reseller] = uint256(j + 1);
                         break;
                     }
                 }
                 else if(!takenT3Chunks.contains(cId)) { // This is a T3, find a chunk that isn't claimed as T3
                     assignedChunkid = cId;
-                     reservedChunksCounter[reseller] = uint256(i + 1);
+                     reservedChunksCounter[reseller] = uint256(j + 1);
                     break;
                 }
                 
--- a/contracts/NFTStorage.sol
+++ b/contracts/NFTStorage.sol
@@ -537,7 +537,7 @@ contract NFTStorage is IERC721ReceiverUpgradeable, Initializable, AccessControlU
     }
 
     // Takes bridgedNFT not meta to avoid stack too deep
-    function _withdrawWeaponFromBridge(uint256 bridgedNFT, uint256 chainId, uint256 sourceId) internal returns (uint256 mintedId) {
+    function _withdrawWeaponFromBridge(uint256 bridgedNFT, uint256 /*chainId*/, uint256 /*sourceId*/) internal returns (uint256 mintedId) {
         // Get any existing mint that has this global id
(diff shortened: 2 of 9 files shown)


contracts/CBKLandSale.sol
```diff
@@ -745,18 +745,18 @@ contract CBKLandSale is Initializable, AccessControlUpgradeable {
             require(reservedTier == TIER_TWO || reservedTier == TIER_THREE, "NA");
 
             uint256 counter = reservedChunksCounter[reseller];
-            for(uint256 i = counter; i < counter + rcLength; i++) {
-                uint256 cId = reservedChunks[reseller].at(uint256(i % rcLength));
+            for(uint256 j = counter; j < counter + rcLength; j++) {
```
kcper, line 748: Are you sure it's going to work? Obviously your change makes sense, but it must have been working earlier I guess and this change looks like it would change the flow. @bilbolPrime is this something you worked on?

contracts/PvpArena.sol
```diff
@@ -526,7 +526,7 @@ contract PvpArena is Initializable, AccessControlUpgradeable {
         bool attackerWon;
     }
 
-    function createDuelist(uint256 id) internal returns (Duelist memory duelist) {
+    function createDuelist(uint256 id) internal view returns (Duelist memory duelist) {
```
kcper, line 529: How can this be a viev if it mutates `duelist.basePower`, what am I missing?

contracts/raidBasic.sol
```diff
@@ -35,14 +35,13 @@ contract RaidBasic is Initializable, Raid {
...
         // we may want to have a specific lockout in the future
-        int128 traitMultiplier = 0;
         uint24 power = 0;
         totalPower += power;
         raiders.push(Raider(uint256(msg.sender), characterID, weaponID, power));
         emit RaiderJoined(msg.sender, characterID, weaponID, power);
     }
 
-    function completeRaid(uint256 seed) public override restricted {
+    function completeRaid(uint256 /*seed*/) public override restricted {
```
kcper, line 44: @poshdan can we get rid of it altogether?

contracts/NFTStorage.sol
```diff
@@ -537,7 +537,7 @@ contract NFTStorage is IERC721ReceiverUpgradeable, Initializable, AccessControlU
     }
 
     // Takes bridgedNFT not meta to avoid stack too deep
-    function _withdrawWeaponFromBridge(uint256 bridgedNFT, uint256 chainId, uint256 sourceId) internal returns (uint256 mintedId) {
+    function _withdrawWeaponFromBridge(uint256 bridgedNFT, uint256 /*chainId*/, uint256 /*sourceId*/) internal returns (uint256 mintedId) {
```
kcper, line 540: Why would we keep these params at all? @bilbolPrime any objections to removing them?

kcper (commented):
Some concerns.

contracts/CBKLandSale.sol
brian-man: The compiler warning is saying that you have "two `i`s" where the latter one hides the former one for the scope of the `for` loop.  My change merely gives a different name to the inner one so that it's not as surprising that they are actually two different variables on the stack.

contracts/PvpArena.sol
brian-man: > How can this be a viev if it mutates `duelist.basePower`, what am I missing?

Changing to `view` indicates that the contract data is not changing.  The `duelist` is the return value, not a contract storage member.

contracts/NFTStorage.sol
brian-man: You could make an argument that the values might be used in the future and thus are a good future flexibility thing.

Also, changing them now means revving the bot and making sure it's the transition is carefully rolled out across all chains.  (Or make the bot smart enough to detect which flavor of the method the contract has.)

contracts/CBKLandSale.sol
bilbolPrime: I am good with this change

contracts/NFTStorage.sol
bilbolPrime: I am good with this change and it's seamless for the bot. Let's not do any migrations for this contract though as it's scheduled for a major upgrade (https://github.com/CryptoBlades/cryptoblades/pull/1143).

contracts/raidBasic.sol
poshdan: Would love to. Though not something I'm looking forward to doing any time soon, seeing as it's already deployed, i dont know if there are any pitfalls to removing a contract.

poshdan:
This wasn't a high priority change and so it got ignored unfortunately.
Could be still relevant to make development processes cleaner. Probably there are new warnings now.
As for when we could bother to fit a multi-contract upgrade migration that is a minor developer quality of life change, is anyone's guess.

PR #1215, opened 2022-03-04
Extract experience table data from Characters.sol source to migration js

brian-man:
This change adds a parameter to `Characters.migrate_1ee400a()` to provide the `experienceTable` data, saving approximately 2.14KB (removing 9%) of the contract size.

More specifically, according to `npx truffle run contract-size`:
- `23.68 KB` Before
- `21.54 KB` After

Testing:
- Deploy locally.
- Validate that the `experience-table.js` values are being pushed: with adjusted values and then with the original values.


kcper (approved):
Looks good.

PR #1226, opened 2022-03-05
Refresh dust and other amounts after submit/complete quest

Chriserus:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

(insert description here)

### Testing


poshdan (approved):
approving.
For posterity I wanted to mention we may soon start revamping blockchain calls in the frontend to fire less calls at once - in this case the input and output would end up determining which update/fetch calls we do, rather than a catch-all

Chriserus:
> 

Good point, for now, I'll merge, but in the future we can revisit store entirely

PR #1229, opened 2022-03-07
Feat/pvp  tweaks and bonus rank formulas

Dvidaurre:
* Add win chance formula to commons.sol
* Add bonus Rank formula to commons.sol
* reduce size on pvpContract and handle bonus rank
*  Remove and mute unused formulas


poshdan (approved):
looks good

PR #1233, opened 2022-03-08
Add over-pay protection to incrementQuestProgress()

brian-man:
When submitting NFTs or amounts, the contract does not protect against overpayment.  While the game UI tries to guide the user away from overpaying for a quest, it can still happen due to quick clicking or with slow responsiveness chains.

This chain adds a check for overpayment and report "Too much" as the revert reason when attempting to pay too much.

### Screenshots

Metamask hint when attempting to over-pay:
image

Sample output from Ganache log when attempting on local:
image

### Testing

Using a lesser dust (2 required amount) quest:
1. Attempt to submit 3 or more
=> fail with "Too much"
2. Submit 1, then submit 2 or more
=> fail with "Too much"
3. Submit 1, then submit 1
=> succeed
4. Submit 2
=> succeed

### DIFF
--- a/contracts/SimpleQuests.sol
+++ b/contracts/SimpleQuests.sol
@@ -359,8 +359,9 @@ contract SimpleQuests is Initializable, AccessControlUpgradeable {
     }
 
     function incrementQuestProgress(uint256 characterID, uint256 questID, uint256 progress) private {
-        uint currentProgress = characters.getNftVar(characterID, NFTVAR_SIMPLEQUEST_PROGRESS);
-        characters.setNftVar(characterID, NFTVAR_SIMPLEQUEST_PROGRESS, currentProgress + progress);
+        uint totalProgress = characters.getNftVar(characterID, NFTVAR_SIMPLEQUEST_PROGRESS) + progress;
+        require(totalProgress <= quests[characterQuest[characterID]].requirementAmount, "Too much");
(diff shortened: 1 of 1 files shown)


Chriserus:
Are you _sure_ this can happen due to "fast clicking" or "slow chain", because as far as frontend goes, the modal itself isn't responsible for loading a quest and knows about quest requirement, progress and goal from the moment it's shown. 

Is the case you mentioned actually reported, or were you able to reproduce it? I didn't add such a check to save on contracts size (earlier I had a problem with contract size, now it's slimmer, but if we can save some space, we should)
@brian-man

brian-man:
> Are you _sure_ this can happen due to "fast clicking" or "slow chain", because as far as frontend goes, the modal itself isn't responsible for loading a quest and knows about quest requirement, progress and goal from the moment it's shown.
> 
> Is the case you mentioned actually reported, or were you able to reproduce it? I didn't add such a check to save on contracts size (earlier I had a problem with contract size, now it's slimmer, but if we can save some space, we should) @brian-man

I reproduced the issue very easily by just choosing 200 stamina for 150 stamina quest.  Neither the frontend nor the contract protect against it without this change.

Originally I encountered the issue because I turned in two sets of 2star weapons for the same quest and got 4/2 before I realized I should have completed earlier.

Chriserus (approved):

PR #1235, opened 2022-03-08
incrementQuestProgress() automatically invokes generateRewardQuestSeed()

brian-man:
When submitting -- either amounts or NFTs -- the `incrementQuestProgress()` function is used to record the quest progress.

This change invokes `generateRewardQuestSeed()` automatically when the quest requirement amount is met, removing the need for the user to confirm (and pay for) an additional transaction.

Note: This change does not improve (or even affect) the `RAID` quest type.

### Screenshots

The `GENERATE REWARD QUEST SEED` version of this confirmation is now unnecessary (for non-`RAID` types):
image

### Testing

Completed about a dozen quests (a mixture of amount and NFT requirements) to confirm that the `GENERATE` confirmation is never offered by Metamask and that the quests complete properly.

No testing done for `RAID` type.


Chriserus:
Seems ok to me, @poshdan do you have anything to add here?

poshdan (approved):
looks good

PR #1237, opened 2022-03-08
Add self-transfer prevention to transferSoul()

brian-man:
While testing the soul transfer UI, I found that the BurningManager contract allows transfer-to-self (which is a no-op).

It seems like that should be reported as an error.

Adding the appropriate `require` triggers an exception in Metamask gas estimation, avoiding wasting the user's gas money.

### Testing

Tested with the soul burn UI, both the success and failure cases.

### DIFF
--- a/contracts/BurningManager.sol
+++ b/contracts/BurningManager.sol
@@ -149,6 +149,7 @@ contract BurningManager is Initializable, AccessControlUpgradeable {
     }
 
     function transferSoul(address targetAddress, uint256 soulAmount) public {
+        require(msg.sender != targetAddress);
         userVars[msg.sender][USERVAR_SOUL_SUPPLY] = userVars[msg.sender][USERVAR_SOUL_SUPPLY].sub(soulAmount);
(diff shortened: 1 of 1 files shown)


poshdan:
why do we need this
Just don't transfer to yourself?

brian-man:
> why do we need this Just don't transfer to yourself?
> Just don't transfer to yourself?

I ran across this issue when testing stuff on local.  Hopping between accounts.  Copy and pasting addresses back and forth from Notepad, etc.

It seemed silly to allow users to waste their gas (and confuse themselves) due to a mispaste, and resulting pointless transaction.

Metamask's UI gives a nice hint -- exception thrown during gas estimation -- to prevent.

Hence offering the PR for consideration as a QoL improvement for players.

poshdan:
This is a thing we can filter on the frontend
At best this saves a tiny amount of gas while adding to the contract size and a tiny gas cost for everyone else who is using the transfer properly.

PR #1248, opened 2022-03-12
Remove unused burnWeapon() method from store.ts

brian-man:


Sellie96 (approved):

poshdan:
PLEASE add at least a short description even if you think the title is self explanatory.
Someone not involved in the relevant files and topics would have no idea what has changed in the framework, what the old behavior was and what the new one is.

PR #1249, opened 2022-03-12
Audit and fix approval logic in store.ts

brian-man:
Users have reported issues with the approval logic failing -- and requiring "surrogate approval" workarounds -- to request the proper approval amounts.

This appears to be caused by improperly accounting for in-game-only (IGO) balances for some contract methods that do not support such balances.

An additional discovery was that many of the frontend approval operations were not accounting for rewards balances when they should have been.  This means requesting more approval than needed or requesting approval when it's unnecessary.

This change does the following:
- renames some of the approval functions for readability and brevity
- adds some new approval functions for convenience and readability
- updates the approval functions used for many operations in store.ts

The approval for these methods now accounts for rewards:
- bridgeItem
- burnCharactersIntoCharacter
- burnCharactersIntoSoul
- createOrder
- orderSpecialWeaponWithSkill
- purchaseCharacterCosmetic
- purchaseCharacterEarthTraitChange
- purchaseCharacterFireTraitChange
- purchaseCharacterLightningTraitChange
- purchaseCharacterWaterTraitChange
- purchaseRenameTag
- purchaseRenameTagDeal
- purchaseShield
- purchaseWeaponCosmetic
- purchaseWeaponRenameTag
- purchaseWeaponRenameTagDeal

The approval for these methods no longer (improperly) accounts for IGO:
- burnWeapons
- joinRaid
- reforgeWeapon
- reforgeWeaponWithDust

### DIFF
--- a/frontend/src/contract-call-utils.ts
+++ b/frontend/src/contract-call-utils.ts
@@ -38,7 +38,7 @@ export async function getFeeInSkillFromUsd(
 
 type WithOptionalFrom<T extends { from: unknown }> = Omit<T, 'from'> & Partial<Pick<T, 'from'>>;
 
-export async function approveFeeFromAnyContractSimple<T extends Contract<unknown>>(
+export async function approveFeeWalletOnly<T extends Contract<unknown>>(
   feeContract: T,
   skillToken: Contracts['SkillToken'],
   from: NonNullable<Web3JsCallOptions['from']>,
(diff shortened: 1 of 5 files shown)


frontend/src/store.ts
```diff
@@ -3408,13 +3411,15 @@ export function createStore(web3: Web3) {
         const { CryptoBlades, SkillToken, Blacksmith } = state.contracts();
         if(!CryptoBlades || !Blacksmith || !state.defaultAccount) return;
 
-        await approveFeeFromAnyContractSimple(
+        await approveFeeWalletOrRewards(
+          CryptoBlades,
```
Sellie96, line 3415: Duplicates on a few of these calls?

brian-man: As discussed in Discord, this is the preexisting usage where one of the params is the rewards holder (CryptoBlades) and the other is the fee spending contract (not always CryptoBlades).

A future refactor could maybe make this simpler/smaller for the "they match" case, but that doesn't work for spenders such as Blacksmith, BurningManager, etc.

Sellie96 (approved):

PR #1251, opened 2022-03-13
Improve gas usage of PvpArena._removeCharacterFromArena()

brian-man:
Replaces `false` assignments in mappings with `delete` operations.


brian-man:
Some context...

https://gastoken.io/
> Writing permanent blockchain state costs a significant amount of gas. For instance, the SSTORE instruction currently costs 20000 gas when writing a non-zero value to storage. Erasing the storage (by overwriting it with zeros) costs an additional 5000 gas, but also provides a refund of 15000 gas.

Issue #1222, opened 2022-03-04
Fix claiming amount estimation


kcper:
The approximation should use full precision of the ratio instead of the displayed one.

In this case the real ratio is 0.00067... SKILL/MONSTA which is 10% higher than what is used in the calculation.
It shows that user will receive more tokens than he really will.
![image](https://user-images.githubusercontent.com/9114200/156846810-f9551ed9-b5ed-4038-9dd1-aa05412d9d5d.png)

PR #1252, opened 2022-03-13
Don't use precision on the ratio in the claim rewards bar, so the claimable value is correct.

Steart:
Before:
Screenshot from 2022-03-13 21-11-41

After:
Screenshot from 2022-03-13 21-09-51

### PR Description

This solves the following issue: https://github.com/CryptoBlades/cryptoblades/issues/1222

The above images show a before and after this change.

### Testing

The code has only been tested by me, but I am not sure if someone needs to check on it.


kcper (approved):
Looks good.

PR #1254, opened 2022-03-14
Fix collision issue between KING and SKILL BasicPriceOracle proxies

brian-man:
### Summary

For local deployments, the `BasicPriceOracle` used for `SpecialWeaponsManager` ends up as the KING oracle rather than the SKILL oracle.  When deploying multiple proxies with the same implementation, the last one "wins".  Migration 92 deploys the KING oracle after migration 8 deploys the SKILL oracle. The result is that all subsequent migrations that attempt to use BasicPriceOracle will use the KING oracle rather than the SKILL oracle.

The solution is to record the proxy addresses by name in a different location in the json and then use that information to reach the desired proxy by name.

### Commits

This change includes the following commits:
- Remove deployProxy import where unused
- Remove upgradeProxy import where unused
- Add deployNamedProxy()
- Migration 8 now uses deployNamedProxy()
- Migration 92 now uses deployNamedProxy()
- Add setNamedProxyAddress()
- Add hasNamedProxy()
- Add 143_repair_basic_oracle_proxies.js
- Add 144_test_basic_oracle_proxies.js
- Add requireNamedProxy()
- Fix oracle in migration 136
- Fix oracle in migration 138

### Testing

- Verified that fresh deployments have both the KING and SKILL oracle addresses in build/contracts/BasicPriceOracle.json
- Verified the "old" deployments (without the fixes to migrations 8, 92, 136, and 138 applied) are repaired by migration 143
- Verified that `SpecialWeaponsManager.priceOracleSkillPerUsd` points to the Skill oracle for both fresh and "old" deployments
- Verified (via migration 144) that both KING and SKILL oracles are directly accessible by future migrations

### DIFF
--- a/migrations/100_nftstorage_logging_for_bot.js
+++ b/migrations/100_nftstorage_logging_for_bot.js
@@ -1,4 +1,4 @@
-const { deployProxy, upgradeProxy } = require('@openzeppelin/truffle-upgrades');
+const { upgradeProxy } = require('@openzeppelin/truffle-upgrades');
 
 const NFTStorage = artifacts.require("NFTStorage");
 
--- a/migrations/101_nftstorage_fee.js
+++ b/migrations/101_nftstorage_fee.js
@@ -1,4 +1,4 @@
-const { deployProxy, upgradeProxy } = require('@openzeppelin/truffle-upgrades');
+const { upgradeProxy } = require('@openzeppelin/truffle-upgrades');
 
 const CryptoBlades = artifacts.require("CryptoBlades");
 const NFTStorage = artifacts.require("NFTStorage");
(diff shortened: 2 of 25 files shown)


brian-man:
This represents at least 12 hours worth of work.

brian-man:
The added json content looks like this:
```
  "networks": {
    "5777": {
      ...
      "links": {},
      "address": "0xDb4c6Eb0dBC7F6a354aCDC9Ae280F5552Fd4f511",
>       "namedProxies": {
>       "SkillPriceOracle": "0x2CF21fBe472f70Bf4B6AC6C548683c7132116B88",
>       "KingPriceOracle": "0xDb4c6Eb0dBC7F6a354aCDC9Ae280F5552Fd4f511"
>     }

```

seiyria:
closing as we no longer use oracles

PR #1258, opened 2022-03-15
Heco giveaway

kcper:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Little contract for Heco special weapon giveaway event (and other giveaways in the future). Address keys in the config needed for Bazaar to use the contract.

### Testing

Locally tested all scenarios.


contracts/partner-giveaways/PartnerGiveaways.sol
```diff
@@ -0,0 +1,87 @@
...
+    function initialize(SpecialWeaponsManager _specialWeaponsManager)
+        public
+        initializer
+    {
+        __AccessControl_init();
+        _setupRole(DEFAULT_ADMIN_ROLE, msg.sender);
+        _setupRole(GAME_ADMIN, msg.sender);
+
+        specialWeaponsManager = _specialWeaponsManager;
+        vars[VAR_SHARDS_REWARD] = 125;
```
Chriserus, line 36: We could move it to the migration file, but it's just one call, so it doesn't matter I guess

contracts/partner-giveaways/PartnerGiveaways.sol
```diff
@@ -0,0 +1,87 @@
...
+        nextGiveawayId = 0;
```
Chriserus, line 37: Do we want to start with 0 here? (we usually start with 1, as 0 is the default value) If 0 is intentional, this assignment is redundant (as it's the default)

Chriserus (approved):

PR #1261, opened 2022-03-16
Fix to incrementQuestProgress()

brian-man:
For users that invoke `generateRewardQuestSeed()` "early", attempts by `incrementQuestProgress()` to do that automatically will fail.

This change makes `incrementQuestProgress()` resilient to that scenario.

This should address the scenario some users are facing where they can successfully submit partial amounts for a quest but when they attempt to submit the required amount it fails.

### Root cause
Users can arrive in the broken state one of two ways directly from the game UI:

Scenario A - Before #1235 

1. Submit a quest (fully)
2. Accept in Metamask
3. Click complete button
4. Accept generate in Metamask
5. Reject complete in Metamask
6. Click skip button
7. Accept in Metamask

Scenario B - After #1235 

1. Submit a quest (fully)
2. Accept in Metamask
3. Click skip button
4. Accept in Metamask


brian-man:
Known characters this should fix:
- Heco 106455

kcper:
I think this was solved in https://github.com/CryptoBlades/cryptoblades/pull/1265

PR #1263, opened 2022-03-17
Make bonus rank formula consistent

Jcalcaldev:
This PR:
- Floors decimal numbers in the winChance and bonusRank formulas, removing some possible differences between them and the solidity-based ones.


package-lock.json
```diff
@@ -17528,72 +17528,72 @@
       "dependencies": {
```
Chriserus, line 17528: I don't think package-lock.json is needed in this PR, as you didn't modify libraries

Chriserus (approved):

PR #1265, opened 2022-03-18
Quest seed fix

poshdan:
An improvement over #1261
It was still possible to request a reward seed early, so I fixed that.
I added a tightly controlled force request call in case we change requirement amounts of quests in the future that would trip up users who have requested a seed but not completed yet by the time the quest input changed.

Credit for the overpay check goes to PR #1233 but due to the seed fix I had to use an altered solution

The code has not been tested yet, but it does compile and deploy fine

### DIFF
--- a/contracts/SafeRandoms.sol
+++ b/contracts/SafeRandoms.sol
@@ -113,11 +113,23 @@ contract SafeRandoms is Initializable, AccessControlUpgradeable {
 
     function requestSingleSeed(address user, uint256 requestID) public restricted {
         _resolveSeedPublic(user);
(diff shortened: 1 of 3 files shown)


kcper (approved):

brian-man:
In #1233:
`require(totalProgress <= quests[characterQuest[characterID]].requirementAmount, "Too much");`

What's the equivalent logic in this PR?

brian-man:
In #1261, this protected against { submit, skip, submit=>fail }:
`if (!hasRandomQuestRewardSeedRequested(characterID)) {`

I'm not seeing how that same protection is present in this PR.

PR #1267, opened 2022-03-21
Add oncePerBlock to Blacksmith.purchaseShield()

brian-man:
This is modeled on the implementation for the mint methods in the game contract (cryptoblades.sol).


poshdan:
Very soon we will be using safe randoms for shields mints too, making this an unnecessary step inbetween that adds a state var we have no use for.
Unfortunately other things took priority at the time of this PR so it slipped.
As far as I'm concerned it's a minor credit.

PR #1279, opened 2022-03-24
Safe transfer from garrison

Chriserus:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Restoring @brian-man PR regarding transferring character directly from garrison. It now has use case (external project interacting with Characters contract)

### Testing


poshdan (approved):
LGTM and I have been told this has been tested

kcper (changes requested):
Looks like there's an issue here, correct me if I'm wrong. @brian-man 
Scenario: User A transfers their character from garrison to User B's garrison.

Watch the state of Garrison as the safeTransferFrom executes:
1. redirectToGarrison
- characterOwner[id] = User B (User B becomes an owner, User A is no longer an owner)

2. redirectFromGarrison
- delete characterOwner[id] (User B is no longer an owner)

The transaction will pass, but it seems that the transferred character will end up with no owner.

Fix would be to replace `if` order in safeTransferFrom so the redirectFromGarrison is executed before redirectToGarrison.

Chriserus:
> 

I changed the order of functions execution, checked the following cases:
garrison to garrison -> ok, restore from garrison -> ok
plaza to plaza -> ok, send to garrison -> ok
garrison to plaza -> ok, send to garrison -> ok
plaza to garrison -> ok, restore from garrison -> ok
garrison to contract (from contract, example, character staking functionality) -> ok

PR #1282, opened 2022-03-25
Add migrations/148_nftmarket_junk_keylootbox_raidtrinket.json

brian-man:
This change adds Junk, KeyLootbox, and RaidTrinket to the allow token types for NFTMarket.


brian-man:
Note that I haven't done any testing for this.  I crafted as quickly as I could because @Nyawmeowww was blocked.

Nyawmeowww:
This is done. Closing now

PR #1293, opened 2022-03-29
Feat/pvp  withdrawal penalty

Jcalcaldev:
This PR:
- Makes it clear for players when they are paying part of their wager to withdraw from the arena.
- Fixes the message in Arena Preparation.
- Displays the withdrawal penalty price in the "Leave Arena" buttons in Arena Summary and Matchmaking. If there is no penalty, a "Free" text is displayed (same as in the re-roll button).


frontend/src/store.ts
```diff
@@ -5435,6 +5435,15 @@ export function createStore(web3: Web3) {
         return reRollFeePercent;
       },
 
+      async getWithdrawFeePercent({ state }) {
+        const { PvpArena } = state.contracts();
+        if (!PvpArena || !state.defaultAccount) return;
+
+        const withdrawFeePercent = await PvpArena.methods.withdrawFeePercent().call({ from: state.defaultAccount });
```
Chriserus, line 5442: You can immediately return instead of assigning to a variable

package-lock.json
```diff
@@ -17528,72 +17528,72 @@
       "dependencies": {
```
Chriserus, line 17528: Package-lock not needed in this pr I think

Chriserus (approved):

Sellie96 (approved):

PR #1303, opened 2022-03-31
Feat/pvp  fix block limit

Jcalcaldev:
This PR:
- Fixes a potential issue with block number limitations when reading from past events.


kcper (approved):
Looks good

PR #1324, opened 2022-04-11
Feat/pvp  split contract

Jcalcaldev:
This PR:

- Splits PvpArena contract in PvpCore and extension contracts. The idea is to plug more extension contracts in the future as we want to add more functionality to the feature. Right now there is only one extension contract: PvpRankings, which handles all the seasonal and rankings related features of PvP.

- Adds an "untiered" arena, where there are no tiers and players can fight players of any level.

- Updates all related tests, front end and migrations.

* Though some contract optimizations were made, there are still more to be done both in terms of gas and contract size. This is the first iteration of the contract splitting.


poshdan (approved):
considering these are new contracts going onto testnet, impact is low (except needing caution about deploying frontend to mainnet), so I'll greenlight this so we can get it tested asap

PR #1329, opened 2022-04-14
+promo banner

kevinunger:
Promo banner 
image

mobile:
image

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,5 +1,10 @@
 <template>
   <div class="app">
+    <Banner
+      text="GIVEAWAY: Win 1 of 10 Level 70+ Characters"
+      link="https://gleam.io/uswdy/cryptoblades-70-character-giveaway"
+      linkText="Enter here"
+    />
     <nav-bar :isToggled="toggleSideBar"/>
     <div class="content dark-bg-text">
       <b-row>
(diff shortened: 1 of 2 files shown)


frontend/src/App.vue
```diff
@@ -1,5 +1,10 @@
 <template>
   <div class="app">
+    <Banner
+      text="GIVEAWAY: Win 1 of 10 Level 70+ Characters"
```
Chriserus, line 4: Wasn't it 71+?

frontend/src/components/Banner.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <div>
+    <div v-if="!hidden" class="banner">
+        <span class="text">{{text}}</span>
+        <a :href="link" class="link" target="_blank" rel="noopener noreferrer">
+          <button class="button" @click="hidden=true">{{linkText}}</button>
+        </a>
+        <svg @click="hidden=true" class="closeBtn" width="27.667" height="27.667" viewBox="0 0 27.667 27.667">
```
Chriserus, line 8: I'm not sure what are the requirements in this task, but are we sure we want it always showing, even after user dismisses it (on component rerender, so, coming to the site for the second time), seems a bit aggressive. We could store the hidden in localStorage and save that user dismissed it, but as I mentioned before, I do not know what the requirements here are

Chriserus (approved):

frontend/src/App.vue
kevinunger: Thats the text Nick sent me. It's also on in the title here: https://gleam.io/uswdy/cryptoblades-70-character-giveaway

Issue #1284, opened 2022-03-25
Refactor: Modularize store


kevinunger:
The current store is massive and needs to be split up.
**Approach 1: Split up the structure**
/store
\- store.ts imports:
-- state.ts
-- getters.ts
-- mutations.ts
-- actions.ts

Actions are currently 4500+ Lines. This can be split up further, if we want that. There are e.g. many calls made only for quests. These can be put into /store/actions/quests.ts
Problems: 
-Splitting it up like that could make the store more complicated than it should be. It's only sorted by actions, getters, mutations and state. E.g. Quests might need actions and state. 

**Approach 2: Vuex Modules**

Example:
![image](https://user-images.githubusercontent.com/5601589/163569898-540415f5-4d90-451e-bbb8-f7f2b69c1f4e.png)

`Store.ts` uses modules e.g. `bridge.ts`. Each module is exactly written like a store, but defined as a normal object instead of a vuex-store. Each module can have it's own state, mutations, actions, getters, but can (and will in most cases) use `rootState` to read the global state and use `commit('...')` / `dispatch('...')` to modify the global state.
This would also allows use to use namespacing. E.g `dispatch('bridge/storeItem')`.

Example:

```
/store/store.ts:

export function createStore(web3: Web3) {
  return new Vuex.Store<IState>({
    modules:{
      bridge,
    },
    state:{ ... } 
    ...
```
 
 
 ```
 /store/bridge/index.ts
 
 import {
  Contract,
  IState,
} from './interfaces';
import {Dispatch} from 'vuex';
import {NftTransfer, TransferedNft} from './interfaces/Nft';

const defaultCallOptions = (rootState:  IState) => ({ from: rootState.defaultAccount });

const bridge = {
  actions: {
    async getNFTChainId(
     {rootState}: {rootState: IState, dispatch: Dispatch}, 
     {nftContractAddr, tokenId}: { nftContractAddr: string, tokenId: number}) {
      const { NFTStorage } = rootState.contracts();
      if(!NFTStorage) return;
      const chainId = await NFTStorage.methods
        .getNFTChainId(nftContractAddr, tokenId)
        .call(defaultCallOptions(rootState));
      return chainId;
    },
  },
};


export default bridge;
 ```

kevinunger:
(approach 2)

- store will be split up into modules that each have their own state, getters, actions, mutations
- they can modify the root state, dispatch root actions and commit root mutations
[...]

PR #1331, opened 2022-04-15
Refactor 1: Issue 1284

kevinunger:
### - issue #1284 (approach 2)

- store will be split up into modules that each have their own state, getters, actions, mutations
- they can modify the root state, dispatch root actions and commit root mutations
- no added namespacing in this first step (can be addded later, by setting `namespacing: true` in the module object)
- vuex modules are simple objects that are written like vuex store

Modules:
- [x] bridge
- [x] pvp
- [x] quests
- [ ] merchandise
- [ ] raid
- [ ] (market)
- [ ] cryptoblades/general
- [ ] staking
- [ ] blacksmith
- [ ] combat
- [ ] SpecialWeaponsManager

Not sure why, but some parts of the current store don't have types like:
https://github.com/CryptoBlades/cryptoblades/blob/64db6bfdb51d39687d92787582767bd6aef3d0e5/frontend/src/store.ts#L3656-L3657

And not sure why ts is not complaining. Need to add them to the new modules.


frontend/src/store/quests/index.ts
```diff
@@ -0,0 +1,622 @@
+import {
+  IState,
+  Contract,
+} from '@/interfaces';
+import {Quest, Rarity, ReputationLevelRequirements, RequirementType, RewardType, TierChances, WeeklyReward} from '@/views/Quests.vue';
+import BigNumber from 'bignumber.js';
+// import Web3 from 'web3';
```
Chriserus, line 7: Missed commented out import

Chriserus (approved):

PR #1339, opened 2022-04-21
Refactor: Store to modules

kevinunger:
### - issue #1284 (approach 2)

- store will be split up into modules that each have their own state, getters, actions, mutations
- they can modify the root state, dispatch root actions and commit root mutations
- no added namespacing in this first step (can be addded later, by setting `namespacing: true` in the module object)
- vuex modules are simple objects that are written like vuex store

To Do:
- [x] merchandise
- [x] remove old market code


kevinunger:
regarding [dbef62c](https://github.com/CryptoBlades/cryptoblades/pull/1339/commits/dbef62c9103aaa4d9dc53b77ccb61ab3ec4ae10c)
Kept the call to check if nft is on the market, get the owner address and the price. Added a btn that could link to bazaar, if bazaar supports query params. referenced in #1341

changed the styling like that (Buy on bazaar btn is disabled):

image
image
image

kevinunger:
removed market, but left the calls to check ownership, price and if it's on sale in.

kcper (approved):
Looks good

PR #1343, opened 2022-04-22
fix enemy card frame

Spintacle123:
PR Includes:
 - Fixed Enemy Image Size
 - Fixed Frame Alignment

### DIFF
--- a/frontend/src/views/Combat.vue
+++ b/frontend/src/views/Combat.vue
@@ -586,7 +586,7 @@ h5{
 }
 
 .encounter img {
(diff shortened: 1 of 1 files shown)


frontend/src/views/Combat.vue
```diff
@@ -629,9 +629,9 @@ h5{
 }
 
 
-.frame-line > img{
+/* .frame-line > img{
```
Chriserus, line 632: Delete it if commented

Chriserus (approved):

PR #1349, opened 2022-04-26
Charge native token fee on combat

Jcalcaldev:
This PR:

charges player a 25% fee on expected earnings when entering combat. This fee is converted to each chain's native token.
This PR will require a bot to be setup (bot is finished and deployed in Heroku. Env vars must be set correctly.) which sets native token's price and skill token's price in USD.

The bot sets prices every one hour. For the feature to work without any issues, we must time the bot so that it sets the prices during the combat hourly rollover (when fights are paused).


contracts/TokensReceiver.sol
```diff
@@ -0,0 +1,39 @@
...
+        address gameContract
+    ) public initializer {
+        __AccessControl_init_unchained();
+        _setupRole(DEFAULT_ADMIN_ROLE, msg.sender);
+
+        game = CryptoBlades(gameContract);
+    }
+
+    function fight(uint256 char, uint256 wep, uint32 target, uint8 fightMultiplier) external payable {
+        uint256 offset = game.fight(char, wep, target, fightMultiplier);
```
kcper, line 31: Was this a contract size issue that it couldn't be implemented in game contract?
Having an external call on fight will up the transaction costs and make it harder to profit.

migrations/164_cb_upgrade.js
```diff
@@ -0,0 +1,16 @@
+const { upgradeProxy } = require("@openzeppelin/truffle-upgrades");
```
kcper, line 1: Migration 163 won't work, because game contract needs upgrading before `setTokensPricesAddress` can be called.
We should have this upgrade in migration 163.

kcper (commented):
Few things to think about:

- Either we revive our BasicPriceOracle as it's not being updated currently and the price is fixed at $62. Reviving it will affect nft minting prices as they are based skill price from BasicPriceOracle.
- External calls and gas costs - for more expensive chains like BSC it will be a big deal, iirc there was a lot of work put into optimizing `fight()` and it seems this could increase the costs by a lot. We are slowly moving away from expensive chains though and it's probably not a big deal on chains like OEC or HECO, so maybe we shouldn't worry in the long run.
- It would be helpful to see the gas costs of `fight()` before and after

contracts/TokensReceiver.sol
Jcalcaldev: Yes, previous contract size was around 23.3 kb, marking the function as payable increased it to around 24.8, which i wasn't able to trim down.

contracts/cryptoblades.sol
```diff
@@ -343,25 +351,29 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
         xpRewards[char] += xp;
 
         vars[VAR_HOURLY_FIGHTS] += fightMultiplier;
         vars[VAR_HOURLY_POWER_SUM] += playerFightPower * fightMultiplier;
 
-        emit FightOutcome(msg.sender, char, wep, (targetPower | ((uint32(traitsCWE) << 8) & 0xFF000000)), playerRoll, monsterRoll, xp, tokens);
+        emit FightOutcome(tx.origin, char, wep, (targetPower | ((uint32(traitsCWE) << 8) & 0xFF000000)), playerRoll, monsterRoll, xp, tokens);
+
+        return (tokens.mul(priceOracleSkillPerUsd.currentPrice()).mul(combatTokenChargePercent))
+            .div(tokensPrices.tokenPrice().mul(10**18));
```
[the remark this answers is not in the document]
Jcalcaldev: Will do

migrations/8_main_game.js
```diff
@@ -46,6 +47,14 @@ module.exports = async function (deployer, network) {
 
   const priceOracle = await deployProxy(BasicPriceOracle, [], { deployer });
 
+  // === This is for development, migration already deployed in other networks ===
```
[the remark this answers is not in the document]
Jcalcaldev: Agreed

frontend/src/data/priceOracleAddresses.json
```diff
@@ -0,0 +1,32 @@
...
+      "name": "hecomainnet",
+      "address": ""
+    },
+    "137": {
+      "name": "polygonmainnet",
+      "address": ""
+    },
+    "1337": {
+      "name": "truffle development",
+      "address": "0x0B7108B29ad73097cF7E549D542915348d885e5f"
```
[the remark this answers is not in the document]
Jcalcaldev: Correction: Removing basic price oracle and setting skill price with the same bot on tokensPrices will allow us to discard this code.

contracts/cryptoblades.sol
```diff
@@ -343,25 +351,29 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
         uint16 xp = getXpGainForFight(playerFightPower, targetPower) * fightMultiplier;
         uint256 tokens = getTokenGainForFight(targetPower, true) * fightMultiplier;
 
-        if(playerRoll < monsterRoll) {
+        uint256 offset = (tokens.mul(combatTokenChargePercent).mul(tokensPrices.skillTokenPrice())).div(tokensPrices.tokenPrice());
```
kcper, line 354: Still 2 external calls here. You can make it only one by creating a function in tokensPrices that will return skillTokenPrice()/tokenPrice() ratio.

contracts/cryptoblades.sol
```diff
@@ -296,17 +302,17 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
     }
 
     function fight(uint256 char, uint256 wep, uint32 target, uint8 fightMultiplier) external
-        onlyNonContract() {
+        onlyNonContract() returns (uint256) {
```
poshdan, line 305: the onlyNonContract modifier no longer makes sense here and users can bypass the fee by just calling this method directly. Should be a restricted check instead

poshdan (changes requested):
left some notes

brian-man:
> priceOracleAddresses.json must be updated ONCE with the SKILL ORACLE addresses. I could not obtain these addresses from the abi due to migration 92 replacing it with the KING ORACLE's one.

https://github.com/CryptoBlades/cryptoblades/pull/1254

contracts/TokensManager.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    function getSkillToNativeRatio() public view returns (int128) {
+        return ABDKMath64x64.divu(skillTokenPrice, tokenPrice);
+    }
+
+    function setTokenPrice(uint256 price) external restricted {
+        tokenPrice = price;
+    }
+
```
poshdan, line 59: A little sidebar here, not necessary to change anything now but, for the future:
The bot that updates the native token and skill token prices might want to send updates for both at once, in which case it would make sense for a function that takes parameters for both. This is only really relevant for cutting costs on expensive networks, and only if the rate of updates is frequent enough to make a difference.

contracts/cryptoblades.sol
```diff
@@ -322,7 +316,7 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
             target,
             now / 1 hours
         );
-        uint256 offset = performFight(
+        uint256 expectedTokens = performFight(
```
poshdan, line 319: technically you could put the return clause here without the need for a variable initialization but this is fine too

migrations/164_cb_upgrade.js
```diff
@@ -1,7 +1,6 @@
 const { upgradeProxy } = require("@openzeppelin/truffle-upgrades");
```
poshdan, line 1: Is this second migration file necessary? I might be missing something here

poshdan (approved):
approved with a few comments

Jcalcaldev:
@poshdan added functionality so that players get refunded upon loss, please review that as well.

This also involves that on deployment, some native token must be transfered to TokensManager (made it payable). This is to avoid first players losing and not being refunded. If most players tend to win, which I believe they do, this issue should be solved almost immediately.

poshdan (approved):

PR #1355, opened 2022-04-28
New UI fixes Alignment issues

Sellie96:
Solved some alignment issues caused by css compensation from previous issues


kevinunger (approved):
lg

PR #1362, opened 2022-05-04
Cb launchpad

kcper:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Details: https://wiki.apps.cryptoblades.io/books/launchpad

Launchpad in one contract - size is 21.46 KiB.
Considering moving vesting/claiming part into another contract for unlocking some space in Launchpad.sol for a price of external call on claiming. I don't expect any more logic to be needed in this contract though and 2.5 KiB is still a decent bit of space left.

Periodic vestings need to be sorted by enabled timestamp in the front-end. Currently it's done by looking up PeriodicVestingEnabled events and figuring out the order with that. I'm not sure if there's a limit to how many blocks back it can look up events from. I might need to add another mapping for launchId -> vestingId -> enabledTimestamp.

### Testing

Tested locally.


contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+            1
+        );
+
+        emit LaunchAdded(nextLaunchId, 1);
+        // nextLaunchId + 1 is reserved for Phase 2s
+        nextLaunchId += 2;
+    }
+
+    function addSecondPhaseForLaunch(uint256 launchId, uint256 startTime) external onlyPhase1(launchId) restricted {
+        Launch memory lp = launches[launchId];
```
bilbolPrime, line 283: Please add a check to validate second phase still not launched

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+            }
+        }
+    }
+
+    function getLaunchAllocationForTier(uint256 launchId, uint256 tier) public view returns (uint256) {
+        return launchBaseAllocation[launchId].mul(tierAllocationWeight[tier]);
+    }
+
+    function getUserRemainingAllocationForLaunch(address user, uint256 launchId) public view returns (uint256 maxAllocation) {
+        if(launchId % 2 == 0) {
```
bilbolPrime, line 144: Wouldn't hurt to validate launch id  > 0 and < next launch unless there's no merit in the check

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+        launchBaseAllocation[launchId + 1] = launchBaseAllocation[launchId];
+
+        emit LaunchAdded(launchId + 1, 2);
+    }
+
+    function removeLaunch(uint256 launchId) external restricted {
+        delete launches[launchId];
+    }
+
+    // UPDATING DETAILS
```
bilbolPrime, line 310: There is no check that the launch id is valid (> 0 < next launch) but should be fine as these are dev only and data can be removed if set incorrectly

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+
+    // WHITELISTING
+
+    function setEligibleUsersData(uint256 launchId, address[] calldata users, uint256[] calldata stakedAmounts) external onlyPhase1(launchId) restricted {
+        require(nextLaunchId >= launchId, "Wrong ID");
+        require(users.length == stakedAmounts.length, "Bad input");
+        require(launchStartTime[launchId] > block.timestamp, "Launch started");
+
+        uint totalWeight = 0;
+        for(uint i = 0; i < users.length; i++) {
```
bilbolPrime, line 403: No check if user length is 0 (which leads to total weight 0).
Also what happens if all staked amounts have "0 tier"? shouldn't we reject certain amounts?

bilbolPrime (changes requested):

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+        skillPrice = price;
+    }
+
+    // USER FUNCTIONS
+
+    function invest(uint256 launchId, uint256 amount) external isWhitelistedFor(launchId) {
+        Launch memory lp = launches[launchId];
+        require(launchStartTime[launchId] != 0 && block.timestamp > launchStartTime[launchId], "Launch not started");
+        require((lp.phase == 1 && block.timestamp < launchStartTime[launchId] + vars[VAR_FUNDING_PERIOD_PHASE_1])
+            || (lp.phase == 2 && block.timestamp < launchStartTime[launchId] + vars[VAR_FUNDING_PERIOD_PHASE_2]), "Launch ended");
```
bilbolPrime, line 449: shouldn't the second one be block.timestamp < launchStartTime[launchId + 1]?

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    // WHITELISTING
+
+    function setEligibleUsersData(uint256 launchId, address[] calldata users, uint256[] calldata stakedAmounts) external onlyPhase1(launchId) restricted {
+        require(nextLaunchId >= launchId, "Wrong ID");
+        require(users.length == stakedAmounts.length, "Bad input");
+        require(launchStartTime[launchId] > block.timestamp, "Launch started");
+
+        uint totalWeight = 0;
```
kcper, line 402: This needs to be assigned with `launchFundsToRaise[launchId].div(launchBaseAllocation[launchId])` to allow for calling this function in batches. I think thousands of entries per array in arguments can hit the gas limits.

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+
+    // VESTING
+
+    function enablePeriodicVesting(uint256 launchId, uint256 percentage) external onlyPhase1(launchId) isTokenAddressSet(launchId) restricted {
+        require(launchLinearVestingsDurations[launchId] == 0, "Linear vesting enabled");
+        require(getTotalUnlockedPercentage(launchId) + percentage <= 100, "Total percentage exceeded");
+        uint256 totalLaunchRaised = launchTotalRaised[launchId].add(launchTotalRaised[launchId + 1]);
+        uint256 contractBalance = IERC20(launchTokenAddress[launchId]).balanceOf(address(this));
+        uint256 decimals = ERC20(launchTokenAddress[launchId]).decimals();
+        if(decimals > 18) {
```
Chriserus, line 363: Seems similar to the method you did in common.sol, maybe we can use it here as well?

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+        emit PeriodicVestingEnabled(launchId, launchPeriodicVestingsPercentages[launchId].length - 1, block.timestamp);
+    }
+
+    function enableLinearVesting(uint256 launchId, uint256 duration) external onlyPhase1(launchId) isTokenAddressSet(launchId) restricted {
+        require(duration > 0);
+        require(launchPeriodicVestingsPercentages[launchId].length == 0, "Periodic vesting enabled");
+        uint256 totalLaunchRaised = launchTotalRaised[launchId].add(launchTotalRaised[launchId + 1]);
+        uint256 contractBalance = IERC20(launchTokenAddress[launchId]).balanceOf(address(this));
+        uint256 decimals = ERC20(launchTokenAddress[launchId]).decimals();
+        if(decimals > 18) {
```
Chriserus, line 382: Same here

contracts/Launchpad.sol
```diff
@@ -0,0 +1,492 @@
...
+        uint256 totalUserInvestment = launchUserInvestment[launchId][msg.sender].add(launchUserInvestment[launchId + 1][msg.sender]);
+        uint256 lastClaimTimestamp = userLinearVestingClaimTimestamp[msg.sender][launchId];
+        if(lastClaimTimestamp == 0) {
+            lastClaimTimestamp = launchLinearVestingsStartTimestamps[launchId];
+        }
+        claimAmount = totalUserInvestment.mul(1e18).div(launchTokenPrice[launchId]).mul(block.timestamp - lastClaimTimestamp).div(launchLinearVestingsDurations[launchId]);
+        claimAmount = Common.adjustDecimals(claimAmount, decimals);
+    }
+
+    function getLaunchBaseInfo(uint256 launchId) public view returns (
```
poshdan, line 172: I believe the launches mapping can be queried directly to return the struct as an array of values.
This can save contract space if needed, but there is no reason to change this otherwise (i think)

Example with quests struct:
frontend/src/store/quests/index.ts line 47
https://github.com/CryptoBlades/cryptoblades/blob/main/frontend/src/store/quests/index.ts
referencing SimpleQuests.sol struct mapping at line 94
https://github.com/CryptoBlades/cryptoblades/blob/main/contracts/SimpleQuests.sol

contracts/Launchpad.sol
kcper: I think there's no point in checking that.

contracts/Launchpad.sol
kcper: Added the check. Nothing happens if they are tier 0, effect is the same as if we rejected these amounts.

contracts/Launchpad.sol
kcper: No, invest function can be called on phase 2 ids, so if it's phase 2 it will already have correct id - no need to adjust with + 1.

contracts/Launchpad.sol
kcper: I think it's fine, saves a bit on contract size to skip the check.

bilbolPrime (approved):

PR #1364, opened 2022-05-06
Contract addresses from json

kevinunger:
Read addresses from /build/contracts/... 
Removes calls to read addresses

### DIFF
--- a/frontend/src/contracts.ts
+++ b/frontend/src/contracts.ts
@@ -12,40 +12,43 @@ import { networks as lp2TokenNetworks } from '../../build/contracts/ExperimentTo
 import { networks as cbkLandT1StakingRewardsNetworks } from '../../build/contracts/CBKLandT1StakingRewardsUpgradeable.json';
 import { networks as cbkLandT2StakingRewardsNetworks } from '../../build/contracts/CBKLandT2StakingRewardsUpgradeable.json';
 import { networks as cbkLandT3StakingRewardsNetworks } from '../../build/contracts/CBKLandT3StakingRewardsUpgradeable.json';
-import { networks as cbkLandNetworks } from '../../build/contracts/CBKLand.json';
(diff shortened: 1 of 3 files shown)


frontend/src/contracts.ts
```diff
@@ -230,88 +233,69 @@ async function setUpStakingContracts(web3: Web3) {
...
-  }
-  catch(e) {
-    console.error(e);
-  }
-
-  if(specialWeaponsManagerAddr) {
-    SpecialWeaponsManager = new web3.eth.Contract(specialWeaponsManagerAbi as Abi, specialWeaponsManagerAddr);
-  }
+  const specialWeaponsManagerAddr = (specialWeaponsManagerNetworks as Networks)[networkId]!.address;
+  const SpecialWeaponsManager = new web3.eth.Contract(specialWeaponsManagerAbi as Abi, specialWeaponsManagerAddr);
```
Chriserus, line 253: What will happen when specialWeaponsManagerAddr is not there? From what I see, the previous solution guards it from throwing an error. We need to have it bulletproof, because iirc the missing address and creating a contract from it made app unusable

frontend/src/contracts.ts
```diff
@@ -351,7 +331,7 @@ export async function setUpContracts(web3: Web3): Promise<Contracts> {
     const simpleQuests = new web3.eth.Contract(simpleQuestsAbi as Abi, simpleQuestsContractAddr);
     questsContracts.SimpleQuests = simpleQuests;
     if (simpleQuests.methods.partnerVault) {
```
Chriserus, line 333: This if is checking if partnerVault property is in SimpleQuests contract in order to get the address from there. Now it's not needed I think

kevinunger: removed

kcper (approved):
Looks good

PR #1376, opened 2022-05-13
check if chain is support on load

kevinunger:
- check if chain is support on load

### DIFF
--- a/frontend/src/contracts.ts
+++ b/frontend/src/contracts.ts
@@ -108,7 +108,10 @@ export function getConfigValue(key: string): any {
 
   if(process.env.NODE_ENV === 'development') return '';
   const env = window.location.href.startsWith('https://test') ? 'test' : 'production';
-  const chain = localStorage.getItem('currentChain') || 'BNB';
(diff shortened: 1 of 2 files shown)


frontend/src/contracts.ts
```diff
@@ -108,7 +108,10 @@ export function getConfigValue(key: string): any {
 
   if(process.env.NODE_ENV === 'development') return '';
   const env = window.location.href.startsWith('https://test') ? 'test' : 'production';
-  const chain = localStorage.getItem('currentChain') || 'BNB';
+  let chain = localStorage.getItem('currentChain') || 'BNB';
+  if(!config.supportedChains.includes(chain)){
+    chain = 'BNB';
```
Chriserus, line 113: Should we also maybe set localStorage to a supported chain here?

Chriserus (approved):

frontend/src/contracts.ts
kevinunger: https://github.com/CryptoBlades/cryptoblades/pull/1376/commits/0b5a56061a597206b5a4cdbef739c3d163a4afb1

PR #1380, opened 2022-05-13
add slippage margin to offset

Jcalcaldev:
This PR:
- Adds a slippage margin to the fight offset. This slippage is adjustable.

### DIFF
--- a/migrations/166_tokens_manager_slippage.js
+++ b/migrations/166_tokens_manager_slippage.js
@@ -1,0 +1,8 @@
+const { upgradeProxy } = require("@openzeppelin/truffle-upgrades");
+
+const TokensManager = artifacts.require("TokensManager");
+
+module.exports = async function (deployer, network) {
+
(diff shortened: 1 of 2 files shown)


poshdan:
Is the hourly 30 second downtime for fights not a good enough window to do price updates on?
If it is, the slippage should not be necessary. Unless kroge specifically asked for this. Let me know

migrations/166_tokens_manager_slippage.js
```diff
@@ -0,0 +1,8 @@
+const { upgradeProxy } = require("@openzeppelin/truffle-upgrades");
+
+const TokensManager = artifacts.require("TokensManager");
+
+module.exports = async function (deployer, network) {
+
```
poshdan, line 6: Needs to have an if clause for testnets

poshdan:
I'll get this on testnet now so that testing can be fast tracked

PR #1389, opened 2022-05-18
Fix Heco approvals not initiating

Sellie96:
Reverted forcing gas price during approvals. Although working fine on Okex, HECO is not liking it and the benefits don't outweight the problems caused. Did further testing on normal contract interaction, everything seems to work fine there.


Chriserus (approved):
lgtm

PR #1395, opened 2022-05-18
set base power as level 1 on ffa arena

Jcalcaldev:
This PR:

- Ignores characters' level on base power when they're in FFA arena.


Chriserus (approved):
LGTM but package-lock shouldn't be updated

Issue #1403, opened 2022-05-19
Merchandise code should be removed


kevinunger:
All code related to merchandise should be removed. It moved to bazaar

PR #1456, opened 2022-05-31
rm merchandise code

kevinunger:
- resolves issue #1403


Sellie96:
Just need to remove all the translations for Merchandise too ! :) Rest looks good

PR #1457, opened 2022-05-31
Fix up the dashboard

Sellie96:


kevinunger:
- centered 'join raid' btn on mobile
image

- fixed container height on smaller screens:
image
image
- fixed mobile raid boss
image
image
- fixed overlapping descriptions
image
image

Issue #1447, opened 2022-05-30
Special (pickable) quests


poshdan:
Special quests are event-like offers we present to players when we want to create traffic for certain NFTs and resources.
This extension involves a new pool of quests that can be freely picked to be done by any character regardless of reputation, as well as do quests that are tied only to wallets, without character involvement.

The idea for wallet quests is to allow low barrier of entry options for users migrating from other projects (but possibly other use cases too). All wallet quests can be progressed simultaneously on a single wallet, it doesn't have to be one at a time.

For quests hand picked for characters, it's the usual routine but no reputation requirement, and the user can choose the quest whenever the character is not on a quest already. This serves a similar purpose to wallet quests but we can tie it to character features (such as stamina, xp and rewarding reputation)

Development is already underway (by me), character-pickable quests are done on the "pickableQuests" branch, wallet quests remain to be implemented still.
Expected completion time (before testnet) is a few productive days away (but definitely before June 7th, ideally deploying to testnet by June 3rd to be tested on the weekend.

This will need additional frontend work soon by someone with that skill set, to display the special quests on offer and hook in the buttons to pick them.

PR #1458, opened 2022-06-01
Pickable quests

poshdan:
Adds manually pickable quests and quests for wallets (simultaneous completion on wallet quests allowed)
Full description in issue #1447 

Requires frontend work to show the quests from new tiers (20 for pickable, 30 for wallet) and present buttons that call the new functions to handle them:
submitWalletProgress(uint256 questID, uint256 indexInTier, uint256[] memory tokenIds)
submitWalletProgressAmount(uint256 questID, uint256 indexInTier, uint256 amount)
completeWalletQuest(uint256 questID)
hasRandomWalletQuestRewardSeedRequested(uint256 questID)

We'll have to make sure admins cant create stamina/raid input and experience/non-0 reputation rewards for wallet quests.

indexInTier is the index of the quest in the questTemplates array on tier 30 (the wallet tier), it is used to validate the questID.
This is a weird quirk because we couldn't store "higher than 4" values in quest struct for tier (enum limitation)
The questID is still passed to avoid accidental submissions in case the indices change last second (ie admin deletes a quest while user is submitting, suddenly the user may be submitting items for another quest)
Otherwise we could just send the index and get the ID in-contract. I'm willing to dump that parameter if need be, but this seemed more bulletproof.

Also, character minting is a possible reward now, as requested by Kroge.
I combined a few functions in there too, to save on contract size (rewardQuest function now uses a Reward struct, thus the weekly function could be merged).


Chriserus (approved):
Looks good to me

PR #1476, opened 2022-06-06
UI bug fixes 02

Spintacle123:
PR Includes: Fixed UI Issues
 - 1.3.3) Old placeholder image for rewarded weapon
 - 1.3.5) The weapon icon should be aligned with the char icon (same size + same font color 
 - 1.3.6) 'Claim Rewards' not visibly disabled when no rewards available
 - 1.5.1) Weird behaviour with the icon color; Last one is still colored when selecting another one
 - 1.3.7) After a user selects a char, a weapon and then clicks 'Join Raid' he get's a pop-up that the raid has not yet started. We should just visibly disable the 'Join Raid' Button beforehand, remove the pop-up and add a tooltip on hover over the disabled btn that says 'Raid has not yet started. Please wait xx:xx:xx'; Same when no weapon is selected

### DIFF
--- a/frontend/src/views/Raid.vue
+++ b/frontend/src/views/Raid.vue
@@ -173,7 +173,9 @@
                       </div>
                   </div>
                   <div class="col-lg-12 join-raid">
-                    <button class="claim-btn" @click="promptRewardClaim()" v-tooltip="'Rewards from Previous Raid'">
+                    <button class="claim-btn" :class="rewardIndexes !== null && rewardIndexes.length > 0 ? 'opacity-1': 'opacity-0'"
+                      @click="promptRewardClaim()" v-tooltip="rewardIndexes !== null && rewardIndexes.length > 0 ?
+                      $t('raid.rewardsFromPrevious'): $t('raid.noAvailableRewards')">
                       {{$t('raid.claimRewards').toUpperCase()}}
                     </button>
                     <button v-if="!isMobile()" class="btn-raid"  v-tooltip="$t('raid.joiningCostStamina', {formatStaminaHours})" @click="joinRaidMethod()">
@@ -1733,6 +1735,16 @@ hr.divider {
   margin-left: 1px;
   font-size: 13px;
    /* margin-top: 10px; */
+}
+
+.opacity-0{
+  opacity: 0.4;
+  cursor:not-allowed;
+}
+
+.opacity-1{
+  opacity: 1;
+  cursor: pointer;
 }
 
 .char-info > div:nth-child(2){
--- a/frontend/src/views/Raid.vue
+++ b/frontend/src/views/Raid.vue
@@ -178,7 +178,11 @@
                       $t('raid.rewardsFromPrevious'): $t('raid.noAvailableRewards')">
                       {{$t('raid.claimRewards').toUpperCase()}}
                     </button>
-                    <button v-if="!isMobile()" class="btn-raid"  v-tooltip="$t('raid.joiningCostStamina', {formatStaminaHours})" @click="joinRaidMethod()">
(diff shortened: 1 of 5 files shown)


frontend/src/views/Raid.vue
```diff
@@ -173,10 +173,16 @@
                       </div>
                   </div>
                   <div class="col-lg-12 join-raid">
-                    <button class="claim-btn" @click="promptRewardClaim()" v-tooltip="'Rewards from Previous Raid'">
+                    <button class="claim-btn" :class="rewardIndexes !== null && rewardIndexes.length > 0 ? 'opacity-1': 'opacity-0'"
+                      @click="promptRewardClaim()" v-tooltip="rewardIndexes !== null && rewardIndexes.length > 0 ?
+                      $t('raid.rewardsFromPrevious'): $t('raid.noAvailableRewards')">
```
Sellie96, line 178: This is a wild bit of logic. Can the logic be moved to a Method and the returned boolean be passed on a :disabled=""  there should be a few examples around, you shouldn't need the css then either

frontend/src/views/Raid.vue
```diff
@@ -173,10 +173,16 @@
...
                       {{$t('raid.claimRewards').toUpperCase()}}
                     </button>
-                    <button v-if="!isMobile()" class="btn-raid"  v-tooltip="$t('raid.joiningCostStamina', {formatStaminaHours})" @click="joinRaidMethod()">
+                    <button v-if="!isMobile()" class="btn-raid"
+                    v-tooltip="!isRaidStarted ? $t('raid.errors.raidNotStarted') :
+                    (!this.selectedWeaponId ? $t('raid.errors.selection') : $t('raid.joiningCostStamina', {formatStaminaHours}))"
+                    @click="joinRaidMethod()"
+                    :class="!isRaidStarted || !this.selectedWeaponId ? 'opacity-0': 'opacity-1'">
```
Sellie96, line 185: I believe the same as above here

frontend/src/views/Raid.vue
Spintacle123: copy

> This is a wild bit of logic. Can the logic be moved to a Method and the returned boolean be passed on a :disabled="" there should be a few examples around, you shouldn't need the css then either

noted

Spintacle123: > This is a wild bit of logic. Can the logic be moved to a Method and the returned boolean be passed on a :disabled="" there should be a few examples around, you shouldn't need the css then either

I already tried to add method that trigger the disabled but the opacity is not affected..

Sellie96:
1.3.7) Doesn't seem to be complete here. It's still popping the modal with "Raid hasn't started". Think it wants the tooltip replacing and remaining disabled in this case.

frontend/src/views/Raid.vue
```diff
@@ -172,8 +172,10 @@
                         <nft-icon :isDefault="true" :nft="{ type: '5bdust' }"/>
                       </div>
                   </div>
+                  <!-- :class="rewardIndexes !== null && rewardIndexes.length > 0 ? 'opacity-1': 'opacity-0'" -->
```
Sellie96, line 175: Why is this commented?

frontend/src/views/Raid.vue
```diff
@@ -528,6 +530,10 @@ export default Vue.extend({
       Events.$emit('weapon-inventory', true);
     },
 
+    noRewards(){
+      return this.rewardIndexes === null && this.rewardIndexes.length === 0;
```
Sellie96, line 534: There is a typing issue here. Check the build

Sellie96 (approved):

PR #1482, opened 2022-06-08
Grab reputation tiers

Sellie96:
Remove hardcoded Knight value and grab the tier for the character dynamically


byyyyyyyyy (approved):
looks good

PR #1487, opened 2022-06-09
more memory

kevinunger:
Got this error from the ts worker a lot:
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```
image

 This change increases the memory limit.


Sellie96 (approved):

Sellie96:
Approving for now until we find the cause

PR #1496, opened 2022-06-13
Skeleton loader on staking

Spintacle123:
### All Submissions

image

### New Feature Submissions

### Notes

### PR Description

- Resolve Issue: 1.6.1

### Testing


kevinunger:
Unfortunaly this won't work like that.
The parent component `SelectStakeType` fetches the `stakeOverview` data and passes this to the `StakeSelectorItem`. This happens fairly quick. The longer loading data is the stake detail data fetched with `fetchStakeDetails()` within each child component (in StakeSelectorItem). We need to wait for that data to load until we can hide the loader. 
Currently, you can refresh the page, the skeleton loader is shown, the overview data (stake type name, apy, ...) is then loaded, the loader hides and then the values of totalStakedBalance, walletBallance, etc. are still at 0.

You could check for each value to load and display a placeholder for that within StakeSelectorItem, but your approach is probably better (cloning StakeSelectorItem and modifying the css and just displaying that before data is loaded).

Spintacle123:
> Unfortunaly this won't work like that. The parent component `SelectStakeType` fetches the `stakeOverview` data and passes this to the `StakeSelectorItem`. This happens fairly quick. The longer loading data is the stake detail data fetched with `fetchStakeDetails()` within each child component (in StakeSelectorItem). We need to wait for that data to load until we can hide the loader. Currently, you can refresh the page, the skeleton loader is shown, the overview data (stake type name, apy, ...) is then loaded, the loader hides and then the values of totalStakedBalance, walletBallance, etc. are still at 0.
> 
> You could check for each value to load and display a placeholder for that within StakeSelectorItem, but your approach is probably better (cloning StakeSelectorItem and modifying the css and just displaying that before data is loaded).

updated the PR

Spintacle123:
> > Unfortunaly this won't work like that. The parent component `SelectStakeType` fetches the `stakeOverview` data and passes this to the `StakeSelectorItem`. This happens fairly quick. The longer loading data is the stake detail data fetched with `fetchStakeDetails()` within each child component (in StakeSelectorItem). We need to wait for that data to load until we can hide the loader. Currently, you can refresh the page, the skeleton loader is shown, the overview data (stake type name, apy, ...) is then loaded, the loader hides and then the values of totalStakedBalance, walletBallance, etc. are still at 0.
> > You could check for each value to load and display a placeholder for that within StakeSelectorItem, but your approach is probably better (cloning StakeSelectorItem and modifying the css and just displaying that before data is loaded).
> 
> updated the PR (added isLoading=false after the `fetchStakeDetails()` ) instead of just putting it on mounted.

kevinunger (approved):

PR #1512, opened 2022-06-22
Banners for request skale

Spintacle123:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

(insert description here)

### Testing

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="app">
-    <skale-banner v-if="currentChain == 'BNB'" :walletAddress="this.web3.eth.defaultAccount"/>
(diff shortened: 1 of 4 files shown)


Sellie96:
@Spintacle123 Just for future reference we cannot grab api keys like this in a public repo. Well really any repo tbf. They should be referenced in the env configs then the actual key placed into the config vars within deployment. This is for security and to avoid anybody getting access to the api.

Don't worry about updating the code, I'll take the wrap for anything from here on. Thanks Spin

Spintacle123:
> @Spintacle123 Just for future reference we cannot grab api keys like this in a public repo. Well really any repo tbf. They should be referenced in the env configs then the actual key placed into the config vars within deployment. This is for security and to avoid anybody getting access to the api.
> 
> Don't worry about updating the code, I'll take the wrap for anything from here on. Thanks Spin

ok will do next time.. thanks for the advice

PR #1549, opened 2022-07-22
Fix blacksmith initial load

### LINKED ISSUE #1548: [BUG] - Forgecost loading too long & console error on initial load
### Error on load Blacksmith:1164
image
image
Reason: `forgeCost` is undefined when `skillForgeCost = await ... ` is called.
Reason: state.defaultAccount is null here:
image

kevinunger:
- resolves #1548

updateMintWeaponFee() is called every 2 seconds. Initially, there is an error, because state.defaultAccount isn't set. This means that we get the data after >2s the earliest. 
Polling and not waiting for a state change is bad. Takes long and throws errors.
- removed the whole interval thing
- updateMintWeaponFee() is now updateForgeData() and includes forgeCost, skillForgeCost, reforgeCost, stakedBalanceThatCanBeSpent, fetchSpecialWeapons() etc.
- watching state.defaultAccount from store (using that as a flag in general is a good idea, because this presupposes that contracts are set as well)
- if state.defaultAccount changes from null -> something, we can fetch and display the data safely in <.5s and don't need guards 
- moved the calls from created() to updateForgeData() (same reason)
- updateForgeData() is called after forging as well
- removed console.logs

### DIFF
--- a/frontend/src/views/Blacksmith.vue
+++ b/frontend/src/views/Blacksmith.vue
@@ -752,7 +752,6 @@ export default Vue.extend({
       return this.currentFilteredWeapons.map(w => w.id);
     },
     totalSkillBalance(): BN {
-      console.log(toBN(fromWeiEther(this.skillRewards)).plus(toBN(fromWeiEther(this.inGameOnlyFunds))).plus(toBN(fromWeiEther(this.skillBalance))).toString());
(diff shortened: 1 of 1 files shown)


kevinunger:
tested on test&prod

kcper (approved):
Looks good. Only concern I have is that if there's a spike in prices it wouldn't be reflected in the UI until users switches tabs, account or refreshed the page (pretty unlikely though looking at the historical speed and size of spikes).

PR #1577, opened 2022-08-11
Refactors

jakehein:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Working on Modularizing Store for Combat specific functions/data. This is part of #1284

### Testing


kevinunger:
You are using the getter `getWeaponDurability()` as `combat/getWeaponDurability` in some place, but the getter isn't in the combat module.
I'd say we leave that in the 'general' store and just remove the `combat/` where it is used.

kevinunger:
Some actions/mutations like `updateXpRewards`, `updateCharacterStamina`aren't properly namespaced where they are used. The best, but tedious way to find the all references is to search for all occurences of each function in /store/combat.ts one-by-one in the whole codebase. These also appear as errors or warnings in the console.

frontend/src/views/Combat.vue
```diff
@@ -349,19 +461,19 @@ export default {
...
-    getElementAdvantage(playerElement, enemyElement) {
+    getElementAdvantage(playerElement: number, enemyElement: number) {
       if ((playerElement + 1) % 4 === enemyElement) return 1;
       if ((enemyElement + 1) % 4 === playerElement) return -1;
       return 0;
     },
 
     // for enemy card animaton
-    enter(el, done) {
+    enter(el: any, done: any) {
```
kevinunger, line 476: this could be `el:HTMLElement`

frontend/src/views/Combat.vue
```diff
@@ -370,13 +482,13 @@ export default {
         delay: el.dataset.index * 0.1
       });
     },
-    beforeEnter(el){
+    beforeEnter(el: any){
```
kevinunger, line 485: this could be el:HTMLElement

frontend/src/views/Combat.vue
```diff
@@ -535,7 +649,7 @@ export default {
       }
       return x1 + x2;
     },
```
kevinunger, line 651: We could remove this crazy regex addCommas() and just change L129 to `{{e.power.toLocaleString()}}`, where it is used.

jakehein:
made changes, currently getting issue with state.contracts apparently not being a function. Not sure how I'm getting this error and seemingly nowhere else is. Any input would be appreciated 
image
image

kevinunger:
@jakehein Yeah, you have to use rootState for that. 
The rootState object is the state defined in our store.ts, which has the contracts. The state is your module state of your vuex-module (saw you used that as well already; so I guess that here is just an overlooked typo)

kevinunger (approved):

PR #1578, opened 2022-08-12
SafeRandoms in shields and weapons minting

Chriserus:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Implements SafeRandoms in weapons and shields minting

### Testing


contracts/Blacksmith.sol
```diff
@@ -62,6 +63,7 @@ contract Blacksmith is Initializable, AccessControlUpgradeable {
     mapping(uint256 => address) currencies;
 
     mapping(uint256 => address) public links;
```
poshdan, line 65: Thoughts on trying the links mapping for saferandoms in both files? We can then avoid having the setSafeRandoms function and instead have a "setLink" with a link identifier constant, so we can set future references too without additional functions?

contracts/Blacksmith.sol
```diff
@@ -129,17 +131,32 @@ contract Blacksmith is Initializable, AccessControlUpgradeable {
         IERC20(tokenAddress).safeTransfer(msg.sender, amount);
     }
 
+    function setSafeRandoms(SafeRandoms _safeRandoms) public isAdmin {
+        safeRandoms = _safeRandoms;
+    }
+
+    function hasSeed() public view returns (bool) {
```
poshdan, line 138: May be forward thinking to have a seed ID param on this function for frontend compatibility. Assuming we'll sell more seed-bound things in blacksmith in the future, so the ABI doesnt have to change. The frontend would have to know about the seed id constant though.

frontend/src/store/store.ts
```diff
@@ -1456,6 +1456,10 @@ export default new Vuex.Store<IState>({
           { feeMultiplier: num * 4 * chosenElementFee * slippageMultiplier, allowInGameOnlyFunds: true }
         );
```
poshdan, line 1458: This will need updating to handle the pay first, pop second nature of the flow. Would like to note we will need some visual feedback for the player to know they have a weapon waiting if for whatever reason the payment succeeds but they cancel/fail the claim. This is made more complicated by the fact that the seeds are separated out by amounts. Not sure what's the best approach there. Calling hasSeed 10 times may be a little slow for the RPC to respond, but including that behavior in the solidity function will add to size. I don't have a preference.

Worth mentioning this behavior will have to be updated in the bazaar repo too. (but at least there it's 1 shield at a time)

poshdan (changes requested):

contracts/Blacksmith.sol
Chriserus: That's a nice idea, implemented in both contracts

frontend/src/store/store.ts
```diff
@@ -1432,68 +1432,35 @@ export default new Vuex.Store<IState>({
...
+        } else {
+          await approveFee(
+            CryptoBlades,
+            SkillToken,
+            state.defaultAccount,
+            state.skillRewards,
+            defaultCallOptions(state),
+            defaultCallOptions(state),
+            cryptoBladesMethods => cryptoBladesMethods.getMintWeaponFee(),
+            {feeMultiplier: quantity * chosenElementFee * slippageMultiplier, allowInGameOnlyFunds: true}
```
kevinunger, line 1457: does it still work without the magic number? :-D

Chriserus: Worked locally :p Needs further checking

contracts/cryptoblades.sol
```diff
@@ -517,65 +521,48 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
 
-    function _mintWeaponLogic(uint8 chosenElement, uint256 eventId) internal {
-        //uint256 seed = randoms.getRandomSeed(msg.sender);
-        if(eventId > 0) {
-            specialWeaponsManager.addShards(msg.sender, eventId, 1);
-        }
-        _updateWeaponMintFee(1);
-        weapons.mint(msg.sender, uint256(keccak256(abi.encodePacked(blockhash(block.number - 1), msg.sender))), chosenElement);
+    function hasSeed(uint seedId, uint quantity, uint element) public view returns (bool) {
+        return SafeRandoms(links[LINK_SAFE_RANDOMS]).hasSingleSeedRequest(tx.origin, getSeed(seedId, quantity, element));
```
kcper, line 557: Use msg.sender if there's no reason to use tx.origin

contracts/cryptoblades.sol
```diff
@@ -517,65 +521,48 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
         int128 discountedMintWeaponFee =
             getMintWeaponFee()
                 .mul(PAYMENT_USING_STAKED_SKILL_COST_AFTER_DISCOUNT)
+                .mul(ABDKMath64x64.fromUInt(num))
                 .mul(ABDKMath64x64.fromUInt(chosenElementFee));
         _payContractStakedOnly(msg.sender, usdToSkill(discountedMintWeaponFee));
-
-        _mintWeaponLogic(chosenElement, eventId);
+        _updateWeaponMintFee(num);
+        SafeRandoms(links[LINK_SAFE_RANDOMS]).requestSingleSeed(tx.origin, getSeed(uint(WEAPON_SEED), num, chosenElement));
```
kcper, line 544: Use msg.sender if there's no reason to use tx.origin

contracts/cryptoblades.sol
```diff
@@ -517,65 +521,48 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
             getMintWeaponFee()
-                .mul(PAYMENT_USING_STAKED_SKILL_COST_AFTER_DISCOUNT)
                 .mul(ABDKMath64x64.fromUInt(num))
                 .mul(ABDKMath64x64.fromUInt(chosenElementFee));
-        _payContractStakedOnly(msg.sender, usdToSkill(discountedMintWeaponFee));
-
-        _mintWeaponNLogic(num, chosenElement, eventId);
+        _payContractConvertedSupportingStaked(msg.sender, usdToSkill(mintWeaponFee));
+        _updateWeaponMintFee(num);
+        SafeRandoms(links[LINK_SAFE_RANDOMS]).requestSingleSeed(tx.origin, getSeed(uint(WEAPON_SEED), num, chosenElement));
```
kcper, line 532: Use msg.sender if there's no reason to use tx.origin

kcper (changes requested):
Summary:
- external functions are cheaper to call than public ones
- no strong reasons against tx.origin here other than being inconsistent with current code
- there's no indication for the user about 2-tx minting, at least make the button change text from "forge" to "claim" if they have weapons to claim

contracts/cryptoblades.sol
```diff
@@ -543,22 +548,21 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
     }
 
-    function mintWeapon(uint32 quantity, uint8 chosenElement, uint256 eventId) external onlyNonContract oncePerBlock(msg.sender) {
-        require(quantity > 0 && quantity <= 10);
-        if(eventId > 0) {
-            specialWeaponsManager.addShards(msg.sender, eventId, quantity);
-        }
+    function mintWeapon() external onlyNonContract oncePerBlock(msg.sender) {
+        uint8 chosenElement = uint8((claimWeaponData[msg.sender]) & 0xFF);
+        uint32 quantity = uint32((claimWeaponData[msg.sender] >> 32) & 0xFFFFFFFF);
```
kcper, line 560: I think we should require quantity > 0 here so the transaction fails with an error instead of passing and not minting anything while data is = 0.

frontend/src/store/store.ts
```diff
@@ -1370,18 +1371,52 @@ export default new Vuex.Store<IState>({
...
+
+      const claimWeaponData = +await CryptoBlades.methods.claimWeaponData(state.defaultAccount).call(defaultCallOptions(state));
+      return Number(BigInt(claimWeaponData) >> BigInt(32));
+    },
+
+    async claimWeapons({ state, dispatch }) {
+      const {CryptoBlades} = state.contracts();
+      if (!CryptoBlades || !state.defaultAccount) return;
+
+      await CryptoBlades.methods.mintWeapon().send({from: state.defaultAccount, gasPrice: getGasPrice(),});
```
kcper, line 1393: Extra `,` at the end?

kcper (changes requested):

kcper (commented):
Looks good now

Chriserus:
@poshdan Taloner finished his review, can you please take another look at how we're handling seeds now?

contracts/cryptoblades.sol
```diff
@@ -225,6 +228,8 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
     mapping(address => mapping(uint256 => uint256)) public userVars;
 
     SpecialWeaponsManager public specialWeaponsManager;
+    mapping(uint256 => address) public links;
+    mapping(address => uint256) public claimWeaponData;
```
poshdan, line 232: I'd prefer not to introduce the claimWeaponData variable.
Technically we can call safeRandoms.hasSingleSeedRequested(msg.sender, seed) to check for that.
There's an external function called "encode" on safeRandoms that takes an array to assemble the seed from different pieces, this way we don't have to make the getSeed functions public in cryptoblades/blacksmith.

Let me know if that works

poshdan (changes requested):

contracts/cryptoblades.sol
Chriserus: claimWeaponData wasn't introduced to replace hasSeed or something like that, it was introduced to have knowledge about what weapon (and how many weapons) to mint during mintWeapon call. Thanks to this, we can know in frontend that player has "any" seed and use that knowledge to highlight claim button. Without it we would have to check 10x for the seeds and decide which one is present.

poshdan: Fair point. I'd probably be crazy enough to go for a halfway measure where we called 5 times on saferandoms with a 1-10 iteration for amount in the function each but i'm not gonna send us down that rabbit hole.

But actually I just realised we have a thing called userVars. How about adding a new key constant for claimWeaponData, and using that instead of adding a new mapping?

contracts/cryptoblades.sol
```diff
@@ -517,65 +522,54 @@ contract CryptoBlades is Initializable, AccessControlUpgradeable {
...
-        oncePerBlock(msg.sender)
-    {
-        uint8 chosenElementFee = chosenElement == 100 ? 1 : 2;
-        _payContractConvertedSupportingStaked(msg.sender, usdToSkill(getMintWeaponFee() * num * chosenElementFee));
-        _mintWeaponNLogic(num, chosenElement, eventId);
-    }
-
-    function mintWeapon(uint8 chosenElement, uint256 eventId) external onlyNonContract oncePerBlock(msg.sender) {
+    function generateWeaponSeed(uint32 quantity, uint8 chosenElement, uint256 eventId) external onlyNonContract oncePerBlock(msg.sender) {
+        require(quantity > 0 && quantity <= 10);
```
poshdan, line 526: adding
userVars[msg.sender][USERVAR_CLAIM_WEAPON_DATA] == 0
to the require statement here and in the other generate statement would be user friendly, to prevent accidental loss of funds. But in any case this is fine.

poshdan (approved):

kcper (approved):

PR #1595, opened 2022-08-18
fix 10x forge button skill cost

kevinunger:
Since the UI was redesigned, the forge fee displayed on the 1x mint btn was === fee displayed on the 10x mint btn.

Before:
image
Now:
image

### DIFF
--- a/frontend/src/components/BlacksmithNav.vue
+++ b/frontend/src/components/BlacksmithNav.vue
@@ -37,7 +37,7 @@
           <cb-button class="custom-cb-btn custom-forge-btn gtag-link-others" tagname="weapon_forge_multiple"
             :title="`${disableForge ? $t('blacksmith.coolingForge') : $t('blacksmith.forge')} x10 <br/>`"
             :isLoading="isLoading"
-            :subTitle="`(${forgeCost === '0' ? '0.0000' : forgeCost} SKILL)`" @clickEvent="$emit('onClickForge', 1)"
+            :subTitle="`(${forgeCost === '0' ? '0.0000' : (forgeCost*10).toFixed(4)} SKILL)`" @clickEvent="$emit('onClickForge', 1)"
             :isDisabled="disableX10Forge || disableForge || (disableX10ForgeWithStaked && useStakedForForge)"
             :toolTip="(disableX10Forge) ? $t('blacksmith.disableDynamicMintingForge') : $t('blacksmith.forge10New')"
             :style="disableX10Forge ? 'opacity: 0.5' : ''"


jakehein (commented):
Based on what I'm seeing in Blacksmith.vue on how forgeCost is calculated per individual forge, I believe shifting the decimal place would not be reflective of the actual cost of 10 forges. Is 10x supposed to be a deal for the user (buy 10 at a time and save .05 SKILL or whatever the diff would be), or should this be a shorthand for 10 calls to forge?

kevinunger:
@jakehein The latter.The 10x forge just mints 10 weapons. There isn't a deal.

kevinunger:
But it should be a deal imo

frontend/src/components/BlacksmithNav.vue
```diff
@@ -37,7 +37,7 @@
           <cb-button class="custom-cb-btn custom-forge-btn gtag-link-others" tagname="weapon_forge_multiple"
             :title="`${disableForge ? $t('blacksmith.coolingForge') : $t('blacksmith.forge')} x10 <br/>`"
             :isLoading="isLoading"
-            :subTitle="`(${forgeCost === '0' ? '0.0000' : forgeCost} SKILL)`" @clickEvent="$emit('onClickForge', 1)"
+            :subTitle="`(${forgeCost === '0' ? '0.0000' : (forgeCost*10).toFixed(4)} SKILL)`" @clickEvent="$emit('onClickForge', 1)"
```
jakehein, line 40: After testing, this doesn't seem to accurately reflect the cost of 10 forges. This might actually be a bigger issue, as my tests have produced wildly differing cost values for forging x10 vs individually forging 10 weapons. Either way though, I would suggest calculating the cost of 10 runs of forge and displaying that value, as so we don't catch users off guard with how much gets actually taken out of their wallets.

A breakdown of what I found, starting from the same/close to same starting forge value given of 0.2199:
- If exclusively hitting 'Forge x10', 1 hit gets us to 0.2199, next gets us to 0.2325. This gives us 10 items at the start of the test and 20 at the end with a price difference of 0.0126.
- If exclusively hitting 'Forge x1', 12 hits gets us to 0.2198 for the next forge (consider this the start position for the test). The next 10 hits will yield 0.2198, 0.2197, 0.2209, 0.2223, 0.2238, 0.2252, 0.2265, 0.2281, 0.2295, and 0.2310. The change in price between start and end is 0.0112. This means that it is somehow cheaper to forge individually than it is to forge 10 at a time, with both the cost ratio being lower per forge and the number of items being generated per amount spent being higher.
TL;DR: It is cheaper to forge 22 weapons individually than it is to forge 20 weapons 10 at a time. This seems like an issue. Whether it is in the scope of this project though is up for debate. Either way, the cost to forge isn't constant and thus shouldn't be displayed this way to the user.

jakehein (changes requested):

kcper (approved):

Issue #1592, opened 2022-08-17
Apply new design to weapon context menu


kevinunger:
Context menu still uses the old design.
![image](https://user-images.githubusercontent.com/5601589/185198500-0088f701-0797-4350-b2b2-ace5ec510839.png)

Adjust border-radius, bg-color, text color and font to fit the new design.

PR #1596, opened 2022-08-18
Apply new design to context menu dropdown

kevinunger:
- see #1592 

- Moved the .dropdown-menu class to NftOptionsDropdown and adjusted the css

This shouldn't be unscoped, but should be it's own component.


jakehein (approved):
lgtm

PR #1607, opened 2022-09-01
Background does not fit height

### LINKED ISSUE #1593: Background does not fit height
Quests:
Mounted:
image
Loaded:
image

Bridge:
image

Blacksmith:
image

Leaderboard:
image

Note:
Maybe set display:flex on .app and flex:1 the content. Use a wrapper for the background images on each page. Width: clamp() the content.
There are also some overflow-x issues on most sites. Reason might be the amount of div nesting. Probably can be simplified to make life easier.

jakehein:
### All Submissions

### New Feature Submissions

Closes #1593 

### Notes

### PR Description

Fixing overflow issues across the site and general css issues that arise out of those changes. 

### Testing

Will require additional testing on pages to make sure user experience isn't terrible


frontend/src/views/PlayToEarn.vue
```diff
@@ -242,7 +251,7 @@ h5{
   }
 
   .play-to-earn{
-    height: 70vh;
+    height: -webkit-fill-available;
```
kevinunger, line 254: isn't this just height:100%?

frontend/src/components/smart/PvPArenaPreparation.vue
```diff
@@ -574,6 +574,14 @@ p, li, span {
...
   justify-content: space-between;
 }
+@media  all and (max-width: 1057px) {
+  .mainWrapper{
+    flex-direction: column;
+  }
+  .arenaSignup .bottomList ul li{
+    white-space: inherit !important;
+  }
+}
```
kevinunger, line 584: Can you try adding flex-wrap:wrap to .mainWrapper and setting a min-width / width to the character image? (currently the image just scales down when the screen gets smaller and the wrapper doesn't wrap)
This way we can avoid the media query, which will be invalid if any other component changes

frontend/src/views/Blacksmith.vue
```diff
@@ -1680,10 +1681,10 @@ export default Vue.extend({
 
 .blacksmith-page{
   background-image: url('../assets/blacksmith/cb-reforge-bg.jpg');
-  background-size: cover;
+  background-size: clamp(100%, 100%, 100%) auto;
   background-repeat: no-repeat;
```
kevinunger, line 1685: Maybe cover is better here?
image

frontend/src/views/Blacksmith.vue
```diff
@@ -1680,10 +1681,10 @@ export default Vue.extend({
...
   background-position: top right;
-  min-height: calc(100vh - 120px);
+  //min-height: calc(100vh - 120px);
```
kevinunger, line 1687: rm comment or remove attribute

frontend/src/views/Bridge.vue
```diff
@@ -6,7 +6,7 @@
     <div class="row m-3 justify-content-center">
       <h3>{{$t('bridge.transferNftsToAnotherChain')}}</h3>
     </div>
-    <div class="row mt-3 justify-content-center">
+    <div class="bridge-transfer-fee row mt-3 justify-content-center">
```
kevinunger, line 9: (Optional) This would be a bit better here imo:
```suggestion
    <div class="w-100 row mt-3 justify-content-center">
```

frontend/src/views/Dashboard.vue
```diff
@@ -981,4 +985,7 @@ export default Vue.extend({
     justify-content: center;
   }
 }
+/* .joinRaid{
+    width: max-content;
+  } */
```
kevinunger, line 990: comment

Chriserus (approved):

PR #1634, opened 2022-09-18
Fix refreshes after joining raid, unselect weapon

### LINKED ISSUE #1615: Raid join lacking value refresh
After joining a raid, the skill counter in the top right isn't updated.
Characters and their weapons do seem to update their stamina/durability EVENTUALLY but i suspect it's not directly from the raid join action, but rather a looping update somewhere else. The page also doesn't unselect the character and weapon you've just joined with.

Chriserus:
### All Submissions

### New Feature Submissions

### Notes

### PR Description

Closes #1615

Balances fetch correctly now, same for weapon durability and character stamina. Weapon gets unselected when joined. Minor performance improvements.

### Testing

### DIFF
--- a/frontend/src/views/Raid.vue
+++ b/frontend/src/views/Raid.vue
@@ -669,8 +669,8 @@ export default Vue.extend({
         await this.joinRaid({ characterId: this.currentCharacterId, weaponId: this.selectedWeaponId});
         this.selectedWeaponId = '';
         this.selectedWeapon = null;
-      } catch (e) {
-        console.error(e);
+      } catch (error) {
+        console.error(error);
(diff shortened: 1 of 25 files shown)


frontend/src/views/Raid.vue
```diff
@@ -23,7 +23,7 @@
                     <div class="boss-name">
                       <div>
                         <div class="elements">
-                          <div class="frame-element">.</div>
+                          <div class="frame-element"/>
```
kevinunger, line 26: https://html.spec.whatwg.org/multipage/syntax.html#elements-2

kevinunger (approved):

kevinunger:
approved

PR #1637, opened 2022-09-19
Updated Treasury design

kevinunger:
Current design:
image

With new design:
image
image
image


kevinunger:
+added background (same as the one from blacksmith)
image

kcper (approved):

PR #1643, opened 2022-09-23
Excessive reloading on quest page

### LINKED ISSUE #1639: Excessive reloading on quest page
Since the UI overhaul (i think), the quest page keeps refreshing completely after almost every interaction with a quest: hitting submit, complete and even hitting cancel on a submit modal.
I understand that some users like to multitask, and this could be a way to make it foulproof where otherwise a desync in current network or wallet address could cause inconvenience, but I think a complete refresh is overkill.

How about this: On clicking any quest button, run a quick check on current network & address compared to a cached value in the app, and refresh the page if there is a mismatch. Otherwise do nothing. I'm open to hearing arguments against it or better suggestions, please contact on discord

jakehein:
### All Submissions

### New Feature Submissions

Closes #1639

### Notes

### PR Description

Changed the page reload behavior to only reload individual quest rows instead of the entire page.
Note: quest-wallet quests seem to have bad behavior when the 'complete' button is hit. This is on Main as well. It is unrelated to my changes, and I don't know if this has been addressed already.

### Testing

tested


Chriserus (approved):
LGTM, but please make sure it works with all scenarios (excessive reload was a dirty hack to get all of the scenarios, I just didn't revisit it)
Some examples:
- refreshing filters and chosen nfts in quests that require submitting weapons (open one submission modal, add some nfts, close it, open a different quests, with different requirements, is the modal in a clear state?)
- same as above for "amount" quests (dust, soul etc)
- after quest completion, does the state refresh correctly? as for character quests: new quest shows up, weekly completions get incremented?
- after partial submission, is the state reflected after modal gets closed (after the transaction passes)?
- skipping quests works correctly - new stamina value is shown for newly rolled quest (assuming that the skip was daily free one)
- timers, what happens when the free skip timer goes to 0, or when the weekly reward timer goes to 0? does it refresh correctly?

Issue #1450, opened 2022-05-30
Equipment manager


poshdan:
Equipment manager is going to be a smart contract that will allow users to assign connections between NFTs and generate some precalculated values (mainly power) for use in many other places of the game, thus saving gas without adding constant overhead for each new element.
Design goals include generalizing the inputs (meaning, multiple types of NFTs could potentially go into a particular slot), and storing values efficiently (in a vars mapping) without making commitments to state variables.
In all likelyhood equipment is going to be "versioned", meaning it would require a re-calculation when we adjust the components, thus needing the user to re-run the equip check transaction. (this prevents potentially harmful outcomes when using new values and deprecating old ones)
We might need to put external NFTs into escrow as we cannot put a busy var on them to prevent transfer after equipping. May have an issue if the NFT contract prevents holding more than a certain amount in a single wallet.
It's possible we might need proxy contracts for calculating values per-nft type (similar to nft bridge) so as to avoid size limits on the equipment manager itself.

Example use case:
-user equips weapon, the fight power is calculated for each possible enemy element, and on fighting, only that precalculated value needs to be accessed, making fights cheaper, so long as you don't un/re equip frequently.
-since weapons and all other nfts are attached to the character, we don't need external steps to get values from those. You could have 10 equipped NFTs and to enter a raid or pvp arena, you only need to grab a few numbers from the character vars itself.

Somewhat related #1370

ed3ath:
Things needed to clarify:

1. "You could have 10 equipped NFTs" - Does this mean you could equip 10 nfts of the same type (Ex. 10 weapons) or different?
2. "We might need to put external NFTs into escrow" - Can we implement this with CB's nfts aswell? (Lesser contract size)

poshdan:
After some discussion, a number of things have been finalized, so here's a technical documentation:

-Dynamic number of equipment slots (we should be able to release more slots over time)
-Equipped items will be held in escrow (solves transfer/modification issues without bloating contracts with checks, has the downside that you need to unequip to apply dust on a weapon, for instance)
-Multiple types of NFTs could go into one slot. We'll specify supported addresses per slot (so we could have an external nft equipped as a weapon). And we will specify the method of calculations for that type of address (cb weapons going through weapons contract to get stats, cryptokitties going through our own proxy interface to get some numbers with whatever info we have available
-Stat calculation should never be left to external contracts not governed by the cryptoblades team - it's a security risk if they can upgrade their contract to break the game or add unfair numbers
[...]

PR #1763, opened 2022-11-07
Equipment manager #1450

poshdan:
### PR Description

Implements #1450 
In short: you equip weapon/shield via transaction to precalculate power values, those are then used in features to save gas

### Notes

As of Nov 7th:

Targets are not acquired properly if you refresh the page on the combat tab (only works when navigating to it)
Reading currently equipped item is not handled, equip preview icon not refreshed
Shield equip is missing
Unequipping weapon or shield is not possible from frontend
Some old unused code left over from handling weapons and shields in pvp/raid/combat, especially css
Raid multiplier/power is not read/setup properly, should use the new powerData values (multiplier can be omitted)
+Possibly other things I may have forgotten

### Testing

Tested equipping 1 weapon, doing a couple fights, doing 1 raid. No test on pvp or un/re-equipping

2022-11-07_16-04-55_chrome
2022-11-07_16-05-05_chrome
2022-11-07_16-05-11_chrome
2022-11-07_16-05-16_chrome


Chriserus:
Frontend:
- equipped both weapon and shield
image
- unequipped both weapon and shield
image
- combat unavailable when no weapon equipped
image
- join raid button disabled and info shown when no weapon equipped
image
- enter arena button disabled and info shown when no weapon equipped 
image

It isn't ideal, but it works, we can adjust all the minor things later on. I will also need some requirements when it comes to "when show what" and "how should it look"

contracts/util.sol
```diff
@@ -29,6 +50,16 @@ library RandomUtil {
         return num.sub(tenPercent).add(randomSeededMinMax(0, tenPercent.mul(2), seed));
     }
 
+    function plusMinus10PercentSeededFast(uint256 num, uint256 seed) internal pure returns (uint256) {
```
Chriserus, line 53: Wouldn't it be useful to have percentage as parameter? For potential future use.

contracts/util.sol
```diff
@@ -29,6 +50,16 @@ library RandomUtil {
...
+        uint256 tenPercent = num/10;
+        return num - tenPercent + randomSeededMinMaxFast(0, tenPercent*2, seed);
+    }
+
+    function plusMinus10PercentSeededPrehashed(uint256 num, uint256 hashedSeed) internal pure returns (uint256) {
```
Chriserus, line 58: Same here

contracts/util.sol
poshdan: It would be useful, but I would prefer to cross this bridge another time.
The 10% figure is the most common use case for us right now and we do save a tiny bit of overhead from having to support a percentage parameter.
What i would want to do, if in the future we use other percentage ranges (probably for Journey), is to have a separate function in this file for it.
Contract sizes are only grown by referenced library functions, meaning even if we have tons of functions in here that's not used by most of our files, their size will not grow by extending util.sol unless we actually reference them in the individual contracts.

PR #1823, opened 2022-12-10
Dex liquidity removal option

poshdan:
### Notes

Not something we wanted to have, but as it happens the dex contract didn't work as intended in its first version, so it is paused for the time being, and the liquidity is meant to be moved back to other swap providers until it is fixed.

This PR adds a function that allows withdrawal of the entire liquidity in a pair. It's not included in the admin panel (for now) and must be called manually via console or a scan website.

This is currently hard blocked by equipment manager that's out on testnets but not mainnets, and having to do a manual netlify build to deliver the FE to mainnets, since those automatic builds have been failing for some unknown reason.

EDIT: Updated on Dec 14 to expand functions a bit to allow direct withdrawal, setting liquidity, removing pairs etc

### Testing

Not tested


bilbolPrime (approved):

Chriserus (approved):
LGTM

Alex43ice (approved):

Alex43ice (approved):

Alex43ice (approved):

Issue #1873, opened 2023-02-09
[BUG] - Infinite XP bug for Combat re-appeared


MrBaptista69:
### Prerequisites

- [ ] I have refreshed the chain and am still experiencing the bug
- [ ] I have confirmed I am on the correct chain
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

### Contact Details

Nick

### What does this bug relate to

Design Issues

### Describe the bug

The infinite bug from the combat is back

### Expected behavior

Not having it

### Current behavior

Showing the proper exp

### To Reproduce

At the combat page

### Device

PC

### Browser

Chrome

### Additional Context

![image](https://user-images.githubusercontent.com/11393767/217908818-a07e94fa-8a9d-4ae8-9bfa-09e9f0907fcb.png)

Lesenus:
It's showing after attach weapon to the character and switch to combat without reload the page

PR #1896, opened 2023-03-09
Fix the infinity xp bug by dispatching the fetchPowerData call when updating character weapons.

Steart:
### All Submissions

Before:
image

After:
image

### New Feature Submissions

https://github.com/CryptoBlades/cryptoblades/issues/1873

### Notes

### PR Description

Solves the infinity XP bug in the issue mentioned above. 

The powerData would contain zeros in the following scenario:
1. Recruit a new character.
2. Equip a weapon.
3. Move to the adventure tab (Without in between refreshes).

### Testing

Has been tested by myself, maybe that it would be neater to only dispatch this call when equiping a weapon instead of also dispatching it on unequip calls.

### DIFF
--- a/frontend/src/store/store.ts
+++ b/frontend/src/store/store.ts
@@ -1080,7 +1080,7 @@ export default new Vuex.Store<IState>({
       await dispatch('fetchKeyLootboxes', ownedKeyLootboxIds);
     },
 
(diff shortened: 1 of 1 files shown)


Chriserus (approved):
And it was such a simple change! Good job, LGTM

Issue #1830, opened 2022-12-16
[BUG] - Dusting Issue


MrBaptista69:
### Prerequisites

- [ ] I have refreshed the chain and am still experiencing the bug
- [ ] I have confirmed I am on the correct chain
- [ ] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository

### Contact Details

English

### What does this bug relate to

Design Issues

### Describe the bug

When adding 4B or 5B to a sword has the slider reach 100 instead of the 4B being 25 and 5B being 10, if a users uses 100 4B/5B the sword will receive the correct amount either 25 4B or 10 5B, but the amount over submitted will be lost

### Expected behavior

Slider should reach the max amount for either 4B or 5B so people don't apply more than they need to. 

### Current behavior

If user does anything over 25 4B or 10 5B then the amount that went over will be lost.

### To Reproduce

add more than 25 4B or 10 5B to a weapon and complete transaction

### Device

PC

### Browser

Chrome

PR #1903, opened 2023-03-20
Fix the LB/4B & 5B dusting reforge slider max values.

Steart:
### All Submissions

Before:
image

After:
Screenshot from 2023-03-20 19-28-12

### New Feature Submissions

Closes https://github.com/CryptoBlades/cryptoblades/issues/1830
### Notes

### PR Description

Limit the usable amount of LB/4B & 5B dust when reforging a weapon.
The max button also makes use of the max cap.

The actual issue didn't mention the LB, but the solidity contract shows 100 for LB, so also fixed that so
all three behave "equally".

### Testing

Has been tested by me, test are welcome tho!

### DIFF
--- a/frontend/src/views/Blacksmith.vue
+++ b/frontend/src/views/Blacksmith.vue
@@ -372,11 +372,11 @@
                   </div>
                 </div>
                 <div>
-                  <input v-model="lesserDust" type="range" min="0" :max="getLesserDust()" value="0" steps="1">
+                  <input v-model="lesserDust" type="range" min="0" :max="100" value="0" steps="1">
                 </div>
                 <div>
                   <span>{{lesserDust}}/{{getLesserDust()}}</span>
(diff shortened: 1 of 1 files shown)


frontend/src/views/Blacksmith.vue
```diff
@@ -372,11 +372,11 @@
                   </div>
                 </div>
                 <div>
-                  <input v-model="lesserDust" type="range" min="0" :max="getLesserDust()" value="0" steps="1">
+                  <input v-model="lesserDust" type="range" min="0" :max="100" value="0" steps="1">
```
Chriserus, line 375: Imo it would be nice to extract these "magic numbers" to some constants, even in this file, just for the sake of clean code

Chriserus (approved):

frontend/src/views/Blacksmith.vue
Steart: Good point, I changed it, hopefully this is what you meant!

